PR #103, opened 2024-07-17
Add linux and linux-shell artifacts

CharlyCst:
Those two artifacts are now built in the miralis-artifact-opensbi repository (https://github.com/CharlyCst/miralis-artifact-opensbi). The two artifacts consists in a Linux kernel (built in the CI) packaged as an OpenSBI payload. While we already tested OpenSBI, the previous version embedded a toy kernel that exited directly, whereas the new artifacts boot a fully fledged Linux kernel.

The two new artifacts are:
- linux: a linux kernel that exits as soon as userspace `init` starts executing. Reaching that points requires completely initializing the kernel. This artifact is now run as part of CI.
- linux-shell: the same image but starts an interactive shell instead of exiting when reaching userspace, a nice demo of running Linux on top of Miralis.


CharlyCst:
Note: it turns out Linux fails in the CI. I suspect it is because on older versions of QEMU (6.2.0 in this case) OpenSBI has to touch the `mip` CSR because support for some interrupt delegation might be missing. Unfortunately Miralis doesn't support `mip` yet, so this fails in the CI.

PR #107, opened 2024-07-19
Add timer interrupt integration test

CharlyCst:
This patch extend the interrupt integration test to cover timer interrupts. The test simply read the current time, configure a timer with a deadline, enable interrupts and wait until the interrupt happens. With this test we now exercise the virtual CLINT device and interrupt logic of Miralis.

### DIFF
--- a/firmware/interrupt/src/main.rs
+++ b/firmware/interrupt/src/main.rs
@@ -1,9 +1,9 @@
 #![no_std]
 #![no_main]
 
-use core::arch::asm;
+use core::arch::{asm, global_asm};
 
-use miralis_abi::{setup_firmware, success};
+use miralis_abi::{failure, setup_firmware, success};
 
 setup_firmware!(main);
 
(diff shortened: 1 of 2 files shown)


CharlyCst:
It seems like the new test requires emulation of some store instruction that we do not have support for yet:

```
[Info  | miralis] Hart ID: 0
[Info  | miralis] misa:    0x800000000014112d
[Info  | miralis] DTS address: 0x87000000
    payload: Any { .. },
    message: Some(
        not yet implemented: Instruction not yet implemented: Sd { rs1: X11, rs2: X10, imm: 0 },
    ),
    location: Location {
        file: "src/virt.rs",
        line: 360,
        col: 18,
    },
    can_unwind: true,
    force_no_backtrace: false,
} 
[Info  | miralis::debug] Maximal stack usage: 20696 bytes (63.32%)
```

I can't reproduce it on my machine, so it seems the compiler I have is different and do not generate the same instruction.

PR #109, opened 2024-07-21
Add support for building Miralis for VisionFive2 platform

CharlyCst:
This PR makes it possible to build Miralis for the VisionFive2 platform. To do so it includes the following changes:

- Create a new config for the VisionFive 2 (`./config.visionfive2.toml`).
- Make it possible to define the start address in the config file.
- Add a `just build <path_to_config>` command.

With the combination of all three we can now build Miralis for the VisionFive 2 with `just build config/visionfive2.toml`


CharlyCst:
@fesqvw just tagging you as this might be of interest for what you are doing :)

PR #110, opened 2024-07-22
Refactor metal.rs

NoeTerrier:
There was a lot of asm macros to write and read csr. It's now modularized.
Functions for switching between firmware and payload are now moved to virt.rs as they don't depend on metal any more.
Correction of typos.

This provides better code readability in my opinion.


CharlyCst (approved):
Thanks for the changes :)

PR #111, opened 2024-07-22
Mip can be set with filter

### LINKED ISSUE #105: Add support for `mip` CSR emulation
We noticed that Linux doesn't boot properly on older QEMU versions (see #103), most likely because the CPU doesn't have the same features available as more recent QEMU versions.
While we work toward supporting the VisionFive2 board we might run into the same issues on the hardware, therefore this seems worth investigating.

**How to reproduce**

QEMU can take a `-cpu` argument to specify which CPU model it should use. I managed to reproduce the same failure as seen on the CI on a more recent QEMU version by passing `-cpu sifive-u54` as argument.

To reproduce first run Linux using `just run linux`. This will succeed on recent QEMU versions. Prior to running Miralis with the OpenSBI/Linux payload `just run linux` will print the QEMU command used, something similar to:

```sh
qemu-system-riscv64 --no-reboot -nographic -machine virt -bios .....
```

Copy this command and add `-cpu sifive-u54` at the end, which will trigger the `mip` emulation issue:

```
Boot HART ID              : 0
Boot HART Domain          : root
Boot HART Priv Version    : v1.12
Boot HART Base ISA        : rv64ima
Boot HART ISA Extensions  : zihpm
Boot HART PMP Count       : 8
Boot HART PMP Granularity : 2 bits
Boot HART PMP Address Bits: 54
Boot HART MHPM Info       : 0 (0x00000000)
Boot HART MIDELEG         : 0x0000000000000222
Boot HART MEDELEG         : 0x000000000000b109
    payload: Any { .. },
    message: Some(
        mip emulation is not yet implemented,
    ),
    location: Location {
        file: "src/virt.rs",
        line: 740,
        col: 21,
    },
    can_unwind: true,
    force_no_backtrace: false,
}
[Info  | miralis::debug] Maximal stack usage: 20536 bytes (62.27%)
```

Note: once `mip` is supported we can add Linux back in the CI test workflow.

NoeTerrier:
Close #105 

Allows firmware to set mip with filtered restriction.

### DIFF
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -309,7 +309,13 @@ pub mod mie {
     /// Constant to filter out SIE bits of mstatus
     pub const SIE_FILTER: usize = SSIE_FILTER | STIE_FILTER | SEIE_FILTER | LCOFIE_FILTER;
 
-    // Mstatus fields constants
+    /// Constant to filter out writable bits of mie.
+    pub const MIE_WRITE_FILTER: usize = SIE_FILTER | MSIE_FILTER | MTIE_FILTER | MEIE_FILTER;
+
+    /// Constant to filter out writable bits of mip.
+    pub const MIP_WRITE_FILTER: usize = SSIE_FILTER | STIE_FILTER | SEIE_FILTER;
+
+    // Mie fields constants
     /// SSIE
     pub const SSIE_OFFSET: usize = 1;
     pub const SSIE_FILTER: usize = 0b1 << SSIE_OFFSET;
--- a/firmware/interrupt/main.rs
+++ b/firmware/interrupt/main.rs
@@ -25,7 +25,7 @@ fn test_mie() {
     let res: usize;
     unsafe {
         asm!(
-            "li {0}, 0x42",
+            "li {0}, 0xbbb",
             "csrw mie, {0}",
             "csrr {1}, mie",
             out(reg) _,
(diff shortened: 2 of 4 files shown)


src/arch/mod.rs
```diff
@@ -309,7 +309,13 @@ pub mod mie {
     /// Constant to filter out SIE bits of mstatus
     pub const SIE_FILTER: usize = SSIE_FILTER | STIE_FILTER | SEIE_FILTER | LCOFIE_FILTER;
 
-    // Mstatus fields constants
+    /// Constant to filter out writable bits of mip.
```
CharlyCst, line 312: Seems there is a typo ^^

```suggestion
    /// Constant to filter out writable bits of mie.
```

CharlyCst (approved):
Thanks for the changes :)
There is only a minor typo, I can fix it and merge if that's good for you.

Just to clarify, does this close #105 or not? The PR description is not super clear on that point (it says both closes #105 and that it doesn't).

NoeTerrier:
> Thanks for the changes :) There is only a minor typo, I can fix it and merge if that's good for you.
> 
> Just to clarify, does this close #105 or not? The PR description is not super clear on that point (it says both closes #105 and that it doesn't).

Yes we can close it, sorry I didn't update the description, but the second bug was caused by something else so I did another issue and PR for it (see #115)

PR #113, opened 2024-07-23
First implementation of userspace arch

### LINKED ISSUE #112: Implement Miralis for userspace
It could be useful to have the userspace Architecture implemented. 
It could provide an easy way to test assertion on Miralis behaviours and monitor registers states.

NoeTerrier:
Close #112 

Support for running Miralis in user-space in order to allow unit testing of some of the Miralis logic configuring the CPU state. We expect that this will help enforcing invariants regarding interrupt configuration to prevent breaking them in the future.

Basic userspace architecture function. Simple tests for main.rs and virt.rs. Trap.rs refactoring.

### DIFF
--- a/src/virt.rs
+++ b/src/virt.rs
@@ -48,16 +48,58 @@ pub struct VirtContext {
 }
 
 impl VirtContext {
-    pub fn new(hart_id: usize, nb_pmp: usize) -> Self {
+    pub const fn new(hart_id: usize, nb_pmp: usize) -> Self {
         assert!(nb_pmp <= 64, "Too many PMP registers");
 
         VirtContext {
             host_stack: 0,
-            regs: Default::default(),
-            csr: Default::default(),
+            regs: [0; 32],
+            csr: VirtCsr {
+                misa: 0,
+                mie: 0,
+                mip: 0,
+                mtvec: 0,
+                mscratch: 0,
+                mvendorid: 0,
+                marchid: 0,
+                mimpid: 0,
+                mcycle: 0,
+                minstret: 0,
+                mcountinhibit: 0,
+                mcounteren: 0,
+                menvcfg: 0,
+                mseccfg: 0,
+                mcause: 0,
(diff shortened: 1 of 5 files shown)


src/virt.rs
```diff
@@ -1001,3 +1001,67 @@ where
...
+        let mut mctx = MiralisContext::new(16, hw);
+        let mut ctx = VirtContext::new(0, 16);
+
+        unsafe { Arch::write_csr(Csr::Mideleg, 0x42) };
+
+        unsafe { ctx.switch_from_payload_to_firmware(&mut mctx) }
+
+        assert_eq!(Arch::read_csr(Csr::Mideleg), 0, "Mideleg need to be 0");
+    }
+}
```
CharlyCst, line 1067: It's great to have tests, but we should also think about when we will have to figure out why they fail ^^

As it is not immediately obvious what the tests are doing, could you add a doc comment for each of those functions to express what is the intent of the test? What part are you trying to test, how is the system expected to behave, and what are the expected invariants/properties?

CharlyCst (commented):
Thanks for the changes, it's great to have the user-space environment implemented it is definitely going to be super useful to ensure we don't break Miralis too often :)

I only have two requests regarding documentation before merging this:
- First adding some doc comments to the tests, so that someone who break the tests in the future can quickly understand what they are trying to test, so they can quickly figure out why their code caused an issue.
- The other is for the comment description. In general it is important to understand the changes made to a project, so it is good to describe the intent in the comment message (the content/how is already in the code).
  In this case, the bigger picture is that we are implementing support for running Miralis in user-space in order to allow unit testing of some of the Miralis logic configuring the CPU state. We expect that this will help enforcing invariants regarding interrupt configuration to prevent breaking them in the future. So writing something along those line in the commit message would be super useful :)

src/virt.rs
NoeTerrier: Of course, I'll add this !

PR #114, opened 2024-07-25
Visionfive2 board support

CharlyCst:
This PR adds initial support for VisionFive2 boards.

@Wesdcv pinging you, I created a PR from your branch.

### DIFF
--- a/src/arch/metal.rs
+++ b/src/arch/metal.rs
@@ -408,7 +408,7 @@ unsafe fn find_nb_of_non_zero_pmp(nb_implemented: usize) -> usize {
                 concat!("csrrw {tmp}, pmpaddr", $idx, ", {addr}"),
                 concat!("csrrw {addr}, pmpaddr", $idx, ", {tmp}"),
                 tmp = out(reg) _,
-                addr = inout(reg) 1_usize => read_addr,
+                addr = inout(reg) usize::MAX => read_addr,
                 options(nomem)
             );
             if read_addr == 0 {
--- a/src/arch/metal.rs
+++ b/src/arch/metal.rs
@@ -3,13 +3,11 @@ use core::arch::{asm, global_asm};
 use core::marker::PhantomData;
 use core::{ptr, usize};
 
-use super::{Architecture, MCause, Mode, RegistersCapability, TrapInfo};
-use crate::arch::mstatus::{self, MIE_FILTER, MPP_FILTER};
-use crate::arch::pmp::pmpcfg;
-use crate::arch::{Arch, HardwareCapability, PmpGroup};
+use super::{Architecture, Csr, MCause, Mode, RegistersCapability, TrapInfo};
+use crate::arch::mstatus::{self, MIE_FILTER};
+use crate::arch::{HardwareCapability, PmpGroup};
 use crate::config::PLATFORM_STACK_SIZE;
-use crate::host::MiralisContext;
-use crate::virt::{VirtContext, VirtCsr};
+use crate::virt::VirtContext;
 use crate::{_bss_start, _bss_stop, _stack_start, main};
 
 /// Bare metal RISC-V runtime.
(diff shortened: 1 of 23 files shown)


src/arch/metal.rs
```diff
@@ -621,27 +622,61 @@ _start:
     add t0, t0, t3       // The actual start of our stack
     add t1, t0, t1       // And the end of our stack
 
+    ld t4, __bss_start
+    ld t5, __bss_stop
+
     // Then we fill the stack with a known memory pattern
     li t2, 0x0BADBED0
+
 loop:
```
CharlyCst, line 631: If we have `zero_loop`, maybe we can call this one `stack_loop` for clarity?

src/arch/metal.rs
```diff
@@ -621,27 +622,61 @@ _start:
...
+zero_loop:
+    bgeu t4, t5, verify
+    sb x0, 0(t4)
+    addi t4, t4, 1
+    j zero_loop
+
+verify:
+    ld t4, __bss_start
+    ld t5, __bss_stop
+    j verify_loop
```
CharlyCst, line 646: You don't need that last jump, because the PC will advance to the next instruction which happens to be `verify_loop` after the last instruction of the `verify` block.

```suggestion
```

src/arch/metal.rs
```diff
@@ -621,27 +622,61 @@ _start:
...
+
+verify_loop:
+    bgeu t4, t5, done
+    lb t6, 0(t4)  
+    beqz t6, continue_verification
```
CharlyCst, line 651: It seems there is a logic issue here:
If `t6` is zero, then we go to continue verification. But if `t6` is _not_ zero the branch is not taken and the PC moves to the next instruction, which is also `continue_verification`.
So in other words the verification doesn't verify anything  😅

src/arch/metal.rs
```diff
@@ -621,27 +622,61 @@ _start:
...
+
+continue_verification:
+    addi t4, t4, 1
+    j verify_loop
+
+halt_execution:
+    j halt_execution
```
CharlyCst, line 658: I think is is unreachable in the current implementation.

src/arch/metal.rs
CharlyCst, line 1: Great job at getting that part to work :)

It seems to me that there are some issues in the verify loop, see the comments inline for details.
I am thinking that we might not want that verify loop in assembly here actually... I am anticipating that it will give us headache when we will try to add multi-core support (what if a core modifies a value while another is still doing the verification?). An unfortunately we can't get helpful error messages at this point in the boot process.
So I would suggest to simply remove the verification assembly to keep things simple :)

src/main.rs
```diff
@@ -32,11 +32,13 @@ use crate::virt::{ExecutionMode, VirtContext};
...
 extern "C" {
     pub(crate) static _stack_start: u8;
+    pub(crate) static _bss_start: u8;
+    pub(crate) static _bss_stop: u8;
 }
 
 pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> ! {
     // For now we simply park all the harts other than the boot one
-    if hart_id != 0 {
+    if hart_id != 1 {
```
CharlyCst, line 41: That is going to cause issue on other (non VisionFive2) platforms. For instance on a 1 core QEMU Virt board Miralis will simply stay blocked there.
A solution to fix this issue would be to have a "BOOT_HART_ID" constant on the `Platform` trait. That way the VisionFive2 platform can put 1 there and QEMU Virt 0.

CharlyCst (commented):
Thanks for the great job :)

I pointed out a few details we should fix before merging that, but I thing we are all good for the initial VisionFive2 support once it's done!

src/arch/metal.rs
Wesdcv: yes, there should be something like bnez t6, halt_execution
it probably doesn't matter already though

CharlyCst:
Thanks for the awesome work to get Miralis to boot on the bard!
It's usually the though part, so well done on navigating through this ^^

I'll close in favor of #121, which is the same changes rebased on main.

PR #116, opened 2024-07-26
Fix manual save of register in run_vcpu

### LINKED ISSUE #115: Optimization issue with corrupted state
> We noticed that Linux doesn't boot properly on older QEMU versions (see https://github.com/CharlyCst/miralis/pull/103), most likely because the CPU doesn't have the same features available as more recent QEMU versions.
> While we work toward supporting the VisionFive2 board we might run into the same issues on the hardware, therefore this seems worth investigating.

> 
> How to reproduce
> 
> QEMU can take a -cpu argument to specify which CPU model it should use. I managed to reproduce the same failure as seen on the CI on a more recent QEMU version by passing -cpu sifive-u54 as argument.

In the same setup of issue https://github.com/CharlyCst/miralis/issues/105, after implementing `mip` CSR emulation, we still got a bug. Linux asks opensbi to set a timer, but the overhead of miralis results in the new value for the next timer event to be already in the past. 

Optimization of the code could help to avoid this, (by setting `opt-level = 3` in `[profile.dev]` section of Cargo.toml) but state is then corrupted: 
```
[Info  | miralis] Hart ID: 0
[Info  | miralis] misa:    0x80000000001411ad
[Info  | miralis] DTS address: 0x87e00000
[Error | miralis] Reached maximum number of exits: 2147584848
error: Recipe `run` failed on line 47 with exit code 1
```

NoeTerrier:
Closes #115 

Stack pointer wasn't updated correctly, resulting in wrong save/restoration of manually save registers under full rust compiler optimization.
Add optimization for development.
Slight changes for performance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ platform_visionfive2 = []
 
 [profile.dev]
 panic = "abort"
-opt-level = 0
+opt-level = 3
 
 [profile.release]
 panic = "abort"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ platform_visionfive2 = []
 
 [profile.dev]
 panic = "abort"
+opt-level = 0
 
 [profile.release]
 panic = "abort"
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,14 +118,14 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
         ctx.pc = firmware_addr;
     }
 
-    main_loop(ctx, mctx);
+    main_loop(&mut ctx, &mut mctx);
 }
 
-fn main_loop(mut ctx: VirtContext, mut mctx: MiralisContext) -> ! {
+fn main_loop(ctx: &mut VirtContext, mctx: &mut MiralisContext) -> ! {
     loop {
         unsafe {
-            Arch::run_vcpu(&mut ctx);
-            handle_trap(&mut ctx, &mut mctx);
+            Arch::run_vcpu(ctx);
+            handle_trap(ctx, mctx);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -17,6 +17,7 @@ platform_visionfive2 = []
 
 [profile.dev]
 panic = "abort"
+opt-level = 3
```
CharlyCst, line 20: I am not sure if we should commit to always run with optimization.
What is the current status of Linux? Does it always work with `opt-level = 3`? Is the change to timer frequency necessary?

CharlyCst (commented):
Thanks for the changes!

Let's discuss the `opt-level` case before moving further, I am not sure what is the right direction.

src/main.rs
```diff
@@ -116,14 +116,14 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
         ctx.pc = firmware_addr;
     }
 
-    main_loop(ctx, mctx);
+    main_loop(&mut ctx, &mut mctx);
 }
 
-fn main_loop(mut ctx: VirtContext, mut mctx: MiralisContext) -> ! {
+fn main_loop(ctx: &mut VirtContext, mctx: &mut MiralisContext) -> ! {
```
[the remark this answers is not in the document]
NoeTerrier: While debugging in gdb, I ended up in the `memcopy` function of the arguments copy. I don't think it is necessary to copy here. It's not required for fixing the problem, just nitpick.

Cargo.toml
NoeTerrier: I was thinking about passing the CI by default: we may disable it for dev with other cpu than sifive-u54.

1. It works (linux boots) without changing linux frequency.
2. It doesn't work (linux never boots) with debug mode and the shell eventually freeze and we reach max number of exits. I don't know yet if the frequency change something to this.

PR #123, opened 2024-07-30
Add zephyr to CI

NoeTerrier:
Simple Zephyr firmware that summons two threads, logging in round-robin before exiting. This sample requires a lot of exits as it's not using the cpu a lot.


CharlyCst:
So cool to have Zephyr running in the CA 🙂
That will make sure our interrupts stay functional, thank your for pushing that!

Issue #122, opened 2024-07-30
`hart_id` Variable Mismatch on VisionFive2 Board


Wesdcv:
#### Description
On the VisionFive2 board, the `hart_id` variable does not correspond to the `mhartid` register value after parking the cores. 

#### Steps to Reproduce
1. Check the values for the following code snippet (`main.rs`):
    ```rust
    pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> ! {
        // For now we simply park all the harts other than the boot one
        if hart_id != Plat::get_primary_hart() {
            loop {
                Arch::wfi();
                core::hint::spin_loop();
            }
        }
    }
    ```
2. Observe that `hart_id` gets the value 4, but `Arch::read_csr(Csr::Mhartid)` after the loop returns 1*.
(*) 1 is just the current configured value for `Plat::get_primary_hart()` on VisionFive2 platform.

#### Additional Context
- Commit: `b822f7fe01de20baf3fd850ffa30bc620925a3a4`
- If `Arch::read_csr(Csr::Mhartid)` is read before the loop, it returns 4. However, if read after the loop, it returns 1:
    ```rust
    pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> ! {
      let hart_id = Arch::read_csr(Csr::Mhartid); // here it reads as 4
        if hart_id != Plat::get_primary_hart() { // for some reason the loop is passed correctly only by primary hart
            loop {
                Arch::wfi();
                core::hint::spin_loop();
            }
        }
        let hart_id = Arch::read_csr(Csr::Mhartid); // here it reads as 1
    }
    ```
- By value i mean the output in the logs after the loop: `log::info!("Hart ID: {}", hart_id);`

NoeTerrier:
Closing issues #124 and #164 seems to have removed the problem.

PR #124, opened 2024-07-30
hart_id is getting reassigned so that firmware won't be confused

Wesdcv:
On the VisionFive2 board `hart_id` somehow was turned out to be 4 while `Mhartid` was 1 for the same thread. It's unclear what causes this issue (Refs #122), but as a temporary workaround we can reassign the value after the parking loop.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,6 +50,9 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
             core::hint::spin_loop();
         }
     }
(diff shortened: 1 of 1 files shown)


CharlyCst:
Thanks for the patch, would you mind adding a commit message with basic details so that someone in the future can understand why this was needed?

Ideally that this affects OpenSBI on VisionFive2 and maybe a link to the corresponding issue (https://github.com/CharlyCst/miralis/issues/122). It is often useful when trying to understand why some code was added using git blame.

PR #126, opened 2024-07-31
Visionfive2 build with OpenSBI+U-Boot as virtualized firmware payload

Wesdcv:
In this state u-boot loads fine and it's possible to execute some commands (not thoroughly tested)

### DIFF
--- a/src/platform/visionfive2.rs
+++ b/src/platform/visionfive2.rs
@@ -8,8 +8,10 @@ use spin::Mutex;
 
 use super::Platform;
 use crate::arch::{Arch, Architecture};
+use crate::config::{PLATFORM_NB_HARTS, PLATFORM_STACK_SIZE};
 use crate::device::{self, VirtClint};
 use crate::driver::ClintDriver;
+use crate::{_stack_start, _start_address};
 
 // —————————————————————————— Platform Parameters ——————————————————————————— //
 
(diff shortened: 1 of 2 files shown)


src/platform/visionfive2.rs
```diff
@@ -79,8 +79,8 @@ impl Platform for VisionFive2Platform {
     }
 
     fn get_miralis_memory_start_and_size() -> (usize, usize) {
-        let size = FIRMWARE_START_ADDR - MIRALIS_START_ADDR;
-        (MIRALIS_START_ADDR, size)
+        //let size = FIRMWARE_START_ADDR - MIRALIS_START_ADDR;
+        (MIRALIS_START_ADDR, 0x200000) // Temporary workaround, suggestions appreciated
```
CharlyCst, line 83: In the linker script we have a symbol `_start_address` for the start of Miralis, and a symbol `_stack_start` for the start of the stack. The total size of Miralis is `_stack_start - _start_address + PLATFORM_STACK_SIZE * PLATFORM_NB_HARTS`

src/platform/visionfive2.rs
```diff
@@ -79,8 +81,23 @@ impl Platform for VisionFive2Platform {
...
+                .checked_sub(_start_address as usize)
+                .and_then(|diff| diff.checked_add(PLATFORM_STACK_SIZE * PLATFORM_NB_HARTS))
+                .unwrap();
+        }
+
+        // Find the next power of two greater than or equal to the original size
+        let mut aligned_size = 1;
+        while aligned_size < size {
+            aligned_size <<= 1;
+        }
```
CharlyCst, line 98: Just to let you know (because I find that cool), aligning up to the new power of two can be done in two assembly instructions: an add and a binary and:

```rs
pub fn align_up(addr: usize, align: usize) -> usize {
    (addr + align - 1) & !(align - 1)
}
```

So let's see what happens here:
First the two `align - 1` will be computed at compile time by your compiler if `align` is a constant, so there is no arithmetic operations there. Same thing with the binary not.
The add increases the value by `align - 1` so there are two cases:
- If the value is already aligned it is smaller (by 1) than the next `align`-aligned value.
- Otherwise it becomes bigger or equal than the next `align`-aligned value.
Finally we do a binary and that will align down to `align`. If we take the two cases we had:
- If the value was already aligned, the alignment down puts it back to its original value, so nothing change.
- If the value was _not_ aligned, then it is now bigger or equal than the next `align`-aligned value, so it is rounded down to that value.

In other words: aligning down is super simple, it's a single binary and. So what we do is first make the value bigger and then align it down.

CharlyCst (approved):
Thanks for the changes :)

Because I merged a bunch of stuff in between the PR is not up to date with main, but I'll handle the rebase :)

src/platform/visionfive2.rs
CharlyCst: I just realized that here we don't want to align, but find the next power of two, so I wrote quite a bit on something unrelated, sorry ^^' 

There is a [built-in function](https://doc.rust-lang.org/std/primitive.u64.html#method.next_power_of_two) in rust for that though, which I suppose is optimized too^^

PR #127, opened 2024-07-31
Basic benchmark on time and number of instructions

### LINKED ISSUE #125: Benchmarking Miralis
It could be interesting to have statistics on Miralis. Here is a non-exhaustive list of what could be measures:

- [x] Total number of exits on qemu (for opensbi and Linux)
- [x] Number of exits on the visionfive board (for opensbi and Linux)
- [x] Total number of world switches
- [x] Total number of fimware exits
- [x] Number of instruction for handling a trap in Miralis on the total number of instruction of the firmware/payload. (to benchmark on qemu)
- [x] Benchmark can be run on arbitrary binary
- [x] Latency for handling a trap on the board
- [x] Size of release version of Miralis
- [x] Benchmark init of opensbi
- [x] Benchmark init of Linux

This issue aims to design a way to automate benchmarking. We may want to benchmark on any firmware/payload in the future. 
We may want for example to reduce the number of print from Miralis into the console to reduce latency and get only the relevant part in a file at the end.

NoeTerrier:
Related to #125 

This provides a skeleton for creating new benchmark and an easy way to test part of Miralis code.
Two basic benchmarks are implemented, one measuring the time of execution of some part of the code of Miralis, and the other one the number of retired instructions.
This PR also introduce the `just benchmark` rule to run benchmarks.

The goal is then to have a way to automate computation of statistics using the measurements. 
Maybe to also find a way to not benchmark on all run of `main_loop` to limit overhead of the benchmarking code.


justfile
```diff
@@ -69,3 +70,11 @@ install-toolchain:
 
 # The following line gives highlighting on vim
 # vim: set ft=make :
+
+benchmark firmware=default:
+	mkdir -p {{benchmark_folder}}
+	cargo run --package runner -- run -v --firmware {{firmware}} --benchmark --bench-type=instr\
+	| grep instr | cut -c 59- > {{benchmark_folder}}/instr.bench
```
CharlyCst, line 77: Would those commands work on windows? We don't really support windows but I's good to keep in mind and try to limit the number of platform-specific tools we need to use.

runner/src/main.rs
```diff
@@ -47,6 +47,11 @@ struct RunArgs {
     #[arg(long)]
     /// Path to the configuration file to use
     config: Option<PathBuf>,
+    #[arg(short, long, action)]
+    benchmark: bool,
+    /// Type of benchmark, default is instr
+    #[arg(long, default_value = "instr")]
+    bench_type: String,
```
CharlyCst, line 54: What if we want to measure multiple things at once? Technically RISC-V allows to use multiple counters simultaneously, like instructions retired + elapsed time.

It's probably something we should take into account, it's often useful to have multiple metrics simultaneously when analyzing performances. A typical example would be cache hit + cache misses, both are needed to draw conclusion about the hit ratio, and that is what tools such as `perf` do.

src/main.rs
```diff
@@ -146,10 +149,26 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
 }
 
 fn main_loop(ctx: &mut VirtContext, mctx: &mut MiralisContext) -> ! {
+    #[cfg(any(feature = "benchmark-instr", feature = "benchmark-time"))]
+    let mut bctx = BenchmarkContext::new();
+
     loop {
         unsafe {
+            #[cfg(any(feature = "benchmark-instr", feature = "benchmark-time"))]
```
CharlyCst, line 157: Try to avoid using `cfg`s, they should be used as a last resort because they quickly creep-up and makes Rust tooling much less powerful. Typically the type-checker will not check statements under deactivated `cfg`s, so we could commit code that doesn't compile on the main branch for instance, which is pretty bad.

In terms of performance there is no difference with using an `if` statement with a constant conditions (as we could get using our existing configuration system): the compiler will optimize the dead code away.

CharlyCst (commented):
Some high level comments while the PR is still in progress:

- If we can we should probably rely on the existing configuration mechanism to configure benchmarks. We could add a `[benchmark]` table in toml config for instance.
- We will most likely need to collect multiple metrics at once, we should try to think about how to handle that case.
- Currently there seems to be two separate implementations for the time and for the instructions, could we unify both of those (and support other metrics later on?)

runner/src/main.rs
NoeTerrier: I was thinking to reduce benchmark influence on each other. If we want to measure number of instruction, we don't want to count instruction of the benchmark code. But it will be a fixed amount of difference for each measurement that we can subtract somehow, measuring only the delta, so I can merge everything.

justfile
NoeTerrier: It will not works. I still need to find a way to get the output of Miralis properly. As the benchmark code is running in Miralis, it doesn't have access to std function and cannot write to a file by itself.

I was wondering how we can "wrap" the execution of miralis inside another code that have access to std function. Maybe in order to merge `benchmark.rs` into `benchmark-analyzer` at the end.

CharlyCst: So the way I would do it is to use the runner for this. When you create a command you can redirect the stdin/stdout/stderr however you like. So the runner could capture the stdout stream, filter the lines with the appropriate tag and process them directly from the runner. So basically merging the runner and your new benchmark analyzer.

It seems a process can return an [Output](https://doc.rust-lang.org/std/process/struct.Output.html) in Rust, which collects the stdout as a `Vec<u8>`, maybe that is what you need?

readme.md
```diff
@@ -40,6 +40,14 @@ If `just gdb` can't locate such a binary it will provide a list of supported GDB
...
 The log level can be adjusted using a `config.toml` file. See `./config/example.config.toml` for reference.
 
+## Benchmark
+
+Different benchmark can be run with `just benchmark`.
+
+The default behaviour is counting the number of retired instructions of function `run_vcpu` and `handle_trap` called in `main.rs` with the `default` firmware. You can specify the binary you want to benchmark with `just benchmark <binaries>`. For example, if you want to benchmark `linux` you can run `just benchmark linux`. 
+
+The command will generate text files in a new folder `benchmark-out` with the name of the benchmarks and their output. Then, it will compute and print some statistics.
+
```
CharlyCst, line 50: Great to have some doc! Could you mode it to `docs/tooling.md`? The idea would be to keep the main `readme.md` short and to the point to get people started, and to have more thorough documentation in the doc folder.

runner/src/config.rs
```diff
@@ -25,6 +25,7 @@ pub struct Config {
     pub vcpu: VCpu,
     #[serde(default)]
     pub platform: Platform,
+    pub benchmark: Option<bool>,
```
CharlyCst, line 28: I am thinking that we might need more options to configure the benchmarks later on, so we probably want to create a new table rather than just a single field (for instance to choose what to benchmark).

To that that you would need to create a `Benchmark` struct, add `#[derive(Deserialize, Debug, Default)]`
and `#[serde(deny_unknown_fields)]` and then add one field (for instance `enable` to toggle benchmarks). You can take inspiration from the other tables such as `Log` or `Debug`.

Then in the `config.toml` it will look like that:

```toml
[benchmark]
enable = true
```

And that gives us more room to expend with other fields later on :)

src/arch/metal.rs
```diff
@@ -678,7 +678,6 @@ __bss_stop:
 static STACK_SIZE: usize = PLATFORM_STACK_SIZE;
 
 // ————————————————————————————— Context Switch ————————————————————————————— //
-
```
CharlyCst, line 681: We keep a space between a separator comment and the next line, so please do not delete that line, thanks!

src/benchmark.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub is_running: bool,
+
+    // Temporary value to store previous state (e.g. state when the benchmark started to compare).
+    pub previous_instr_count: usize,
+
+    pub previous_timer: usize,
+}
+
+impl Benchmark {
+    #[allow(dead_code)]
```
CharlyCst, line 21: Hmmm why do you need `#[allow(dead_code)]` here? It seems like it is used?

src/benchmark.rs
```diff
@@ -0,0 +1,74 @@
...
+            asm!(
+                "rdtime {value}", 
+                value = out(reg) stop_time);
+        }
+        Self::record_entry("time", tag, stop_time - self.previous_timer);
+    }
+
+    /// Print formated string with value of the entry and a tag for identification.
+    fn record_entry(bench: &str, tag: &str, entry: usize) {
+        log::debug!("|| {:>15} || {:25} || {}", bench, tag, entry);
```
CharlyCst: if that is easier we could also use a custom print function for benchmarks. That way this would ignore the log level (which is something orthogonal to the benchmarks, and we might even want to benchmark the impact of the log level in the future).

To do that you could create a macro (e.g. `benchmark_print`) that uses `core::format_args!` and calls `Plat::debug_print`. There is an example in the Tyche repo [here](https://github.com/epfl-dcsl/tyche-devel/blob/ec97e692fe5a97d0f0b21cc4818f085e2ca197aa/monitor/tyche/src/debug.rs#L21-L32). If you do that be sure to check if benchmarks are enabled in the macro, so that nothing gets printed at all when benchmarks are disables ^^

src/config.rs
```diff
@@ -211,3 +211,5 @@ pub const PLATFORM_NB_HARTS: usize = {
 /// The stack size for each Miralis thread (one per hart).
 pub const PLATFORM_STACK_SIZE: usize =
     parse_usize_or(option_env!("MIRALIS_PLATFORM_STACK_SIZE"), 0x8000);
+
+pub const BENCHMARK: bool = is_enabled!("BENCHMARK");
```
CharlyCst, line 215: Add a small comment (e.g. `/// Whether benchmarks are enabled`), like for the other constants :)

CharlyCst (commented):
Great work, it's moving forward quite nicely :)

config/example.config.toml
```diff
@@ -2,6 +2,8 @@
 #
 # To configure Miralis create a copy of that file named config.toml
 
+benchmark = false
```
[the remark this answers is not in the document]
NoeTerrier: Sure !

src/arch/metal.rs
NoeTerrier: It's a mistake, I had no intention of changing it haha

src/benchmark.rs
NoeTerrier: It is because I used macro to enable code or not before, so the code wasn't used sometime and this disabled the warning. Now I don't need this anymore, good catch thanks :)

justfile
NoeTerrier: I can use the `output()` function on `qemu_cmd` to collect stdin and stderr and then write it into a file.

I think I'll directly merge benchmarking into the runner so that it can be executed directly from the collected output and avoid use of a file.

Cargo.toml
```diff
@@ -42,7 +42,7 @@ members = [
     "crates/core",
 
     # Tooling
-    "runner",
+    "runner", "benchmark-analyzer",
```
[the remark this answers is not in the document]
NoeTerrier: Yes I was thinking to do that from the beginning, and now I know how to do that :)

But we will still need to take input from a file if we want to benchmark the board.

justfile
CharlyCst, line 1: As you added a new config we also want to test that the config is valid with `just test`. You can add a new line `cargo run -q --package runner -- check-config ./config/qemu-virt-benchmark.toml` toward the beginning of `just test`.

runner/src/config.rs
```diff
@@ -25,6 +25,7 @@ pub struct Config {
     pub vcpu: VCpu,
     #[serde(default)]
     pub platform: Platform,
+    pub benchmark: Benchmark,
```
CharlyCst, line 28: You should add `#[serde(default)]` on the benchmark field, that way the benchmark table is not required in the toml config, which makes it easier to create new configs.

src/benchmark.rs
```diff
@@ -0,0 +1,94 @@
...
+//! the number of instruction for example.
+
+use core::arch::asm;
+
+use crate::arch::{Arch, Architecture, Csr};
+use crate::config;
+use crate::platform::{Plat, Platform};
+
+#[macro_export]
+macro_rules! print {
```
CharlyCst, line 13: We don't really want to expose a normal `print`, so maybe rename it to something that makes it obvious that this macro should not be used by itself (like `_benchmark_print`)

CharlyCst (changes requested):
Thanks for all of the improvements ^^ I think the PR is in an good shape now, I only have a last round of comments (only 3 this time), and then we are good to go :)

PR #128, opened 2024-07-31
Added support for more load&store instructions

Wesdcv:
Overall there are 19 instructions defined in the spec for RV64 (general, compressed, unsigned, floating point)
For now only general&compressed load/store instructions are implemented 
Handler for such instructions also have been refactored to avoid too much repetition, but potentially it could be sized down even more
In the current state in works on OpenSBI and Zephyr on Qemu


src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
+        imm: isize,
+        length: usize,
+    ) {
+        let address = utils::calculate_addr(self.get(rs1), imm);
+        let offset = address - device.start_addr;
+        match device.device_interface.read_device(offset) {
+            Ok(mut value) => {
+                // Sign-extend to 32 bits
+                if length < 32 && value < (1 << 31) {
+                    value = (value as i32) as usize;
```
CharlyCst, line 396: Out of curiosity why is there a 32 bits extend here? Is is mandated by some aspect of the spec or related to our own implementation?

src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
-                let address = self.get(rs1) + imm * 8;
-                let offset = address - device.start_addr;
-                match device.device_interface.write_device(offset, self.get(rs2)) {
-                    Ok(()) => self.pc += 2,
-                    Err(err) => panic!("Error writing {}: {}", device.name, err),
+            Err(err) => panic!("Error writing {}: {}", device.name, err),
+        }
+    }
+
+    fn handle_load(
```
CharlyCst, line 382: Add a bit of doc here too :)

src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
-                let address = self.get(rs1) + imm * 4;
-                let offset = address - device.start_addr;
-                match device.device_interface.write_device(offset, self.get(rs2)) {
-                    Ok(()) => self.pc += 2,
-                    Err(err) => panic!("Error writing {}: {}", device.name, err),
-                }
+            Err(err) => panic!("Error reading {}: {}", device.name, err),
+        }
+    }
+    fn handle_store(
```
CharlyCst, line 412: Add a space in between the two functions, and a small doc comment too ^^

CharlyCst (changes requested):
Thanks for the changes! I mostly have three points:

- Some minor stylistic issues
- Adding unit tests for the decoder
- And a question regarding the handling of `write_device`, to be sure everything is working as expected.

Also, I don't expect this to fit in this PR, but in the long term it would be cool to think about how we could test all of this device code. Maybe we could create a `TestDevice` with some pre-defined value and use it with integration tests? Just wondering :)

src/virt.rs
Wesdcv: From the spec (v2.2, p.19):

> LH loads a 16-bit value from memory, then sign-extends to 32-bits before storing in rd ... LB is defined analogously for 8-bit values

I took it as LH and LB should sign-extend the value. For stores no extension is mentioned.

src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
-                    }
-                    Err(err) => {
-                        panic!("Error reading {}: {}", device.name, err);
-                    }
-                }
+    // TODO: think about passing instructions instead of registers
+    // TODO: maybe join load-stores?
+    // TODO: multiplier = length/8?
+    // TODO: is this the right place for these handlers?
+    fn handle_c_load(
```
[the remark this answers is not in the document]
Wesdcv: Yes, that's compressed
Regarding the documentation - ok, working on it

src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
-                match device.device_interface.read_device(offset) {
-                    Ok(value) => {
-                        self.set(rd, value);
-                        self.pc += 2;
-                    }
-                    Err(err) => {
-                        panic!("Error reading {}: {}", device.name, err);
-                    }
-                }
+    // TODO: think about passing instructions instead of registers
```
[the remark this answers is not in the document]
Wesdcv: Since all load/stores instuctions share registers insinde the same class (load or store), to my mind match won't be needed
Still, there probably isn't much to gain from that change, so registers are fine

src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
+    // TODO: maybe join load-stores?
```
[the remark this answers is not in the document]
Wesdcv: Well, for now 2 separate loads and stores don't look too bad (maybe 30% common code overall and 50-60% between functions of the same class).
But potentially there are also float and unsigned loads and stores for the standard rv64 spec, and we might want to implement them as well.
And if the current pattern continues, it will be 4 loads and 4 stores doing mostly the same, which i believe will reduce code's readability and writability. On the other side, keepeng the handlers separate may ease the process of verification and attesting, as it would be clear what is being done for each type of loads&store and reviewing that against specification is a trivial task.

In other words, i'm unsure if the potential advantages of the join are enough to cover it's shortcomings, so counting on you to determine which approach to take here.

**P.S**: Just to clarify, i meant joining loads and stores among themselves (so like 1 general store and 1 general load ideally)

src/virt.rs
```diff
@@ -314,58 +314,150 @@ impl VirtContext {
...
+    // TODO: multiplier = length/8?
```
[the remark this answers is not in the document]
Wesdcv: 👌

src/virt.rs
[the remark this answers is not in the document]
CharlyCst: It see, that makes sense.
I would say let's keep it like this for now and revisit the decision when we need to add floating points, the current implementation seems clear enough to me.

src/platform/visionfive2.rs
```diff
@@ -79,8 +79,8 @@ impl Platform for VisionFive2Platform {
     }
 
     fn get_miralis_memory_start_and_size() -> (usize, usize) {
-        let size = FIRMWARE_START_ADDR - MIRALIS_START_ADDR;
-        (MIRALIS_START_ADDR, size)
+        //let size = FIRMWARE_START_ADDR - MIRALIS_START_ADDR;
+        (MIRALIS_START_ADDR, 0x200000) // Temporary workaround, suggestions appreciated
```
CharlyCst, line 83: I remember adding a comment here but I can't find it anymore, so I probably discarded it by mistake...

There are symbols for the start of Miralis (`_start_address`) and for the start of the stack (`_stack_start`) The size of Miralis is at `_stack_start - _start_address + PLATFORM_STACK_SIZE * PLATFORM_NB_HARTS`.

You will need to round the address to a power of 2 however, because we use it in the PMP with a NAPOT encoding (which requires sizes that are powers of two).

Wesdcv: I believe that's on another branch, #126

src/device.rs
```diff
@@ -6,6 +6,45 @@ use crate::driver::{clint, ClintDriver};
 
 // ———————————————————————————— Virtual Devices ————————————————————————————— //
 
+// Represents different data widths:
+//  - `Byte`: 8 bits (1 byte)
+//  - `Byte2`: 16 bits (2 bytes)
+//  - `Byte4`: 32 bits (4 bytes)
+//  - `Byte8`: 64 bits (8 bytes)
```
CharlyCst, line 13: Small nit: in Rust there are two types of comments, standard comments with `//`, and documentation comments with `///`.
The difference is that documentation comments (doc comments for short) are used to generate documentation automatically. You can try it on you machine, simply run `cargo doc --open` in the Miralis folder and navigate a bit around :)

The general rule is: when you add comments for documentation purposes (e.g. on top of functions or structures) use doc comments (`///`), that way the documentation can be auto-generated.

src/virt.rs
```diff
@@ -314,44 +314,179 @@ impl VirtContext {
...
-                    Err(err) => {
-                        panic!("Error reading {}: {}", device.name, err);
-                    }
-                }
+    // Handles a compressed load instruction.
+    //
+    // Calculates the memory address and reads a zero-extended value
+    // from the device (after applying a mask to prevent memory leak).
+    // - Compressed load&store instructions are 2 bytes long.
+    // - The immediate (`imm`) value is always positive and has a multiplier applied.
```
CharlyCst, line 322: Similar remark, this would be appropriate for a doc comment :)

CharlyCst (dismissed):
Thanks for the changes! Looks all good to me, I will massage a bit the commits to group related changes and merge into the main branch :)

PR #131, opened 2024-08-07
Refactor benchmark code and add more counters

NoeTerrier:
Code is refactored to make it easier to add a counter. Two type of counters: first one is basic counters yith purpose of counting occurences of events. The second one aims to measure difference beetween two events.
The counters shares many properties that are grouped under the Either enum.

Other properties are recorded like number of world switches or firmware exits.

Related to #125.

### DIFF
--- a/firmware/ecall_benchmark/main.rs
+++ b/firmware/ecall_benchmark/main.rs
@@ -1,0 +1,22 @@
+#![no_std]
+#![no_main]
+
+use core::arch::global_asm;
+
+use miralis_abi::firmware_panic;
+
+// This ecall Miralis with benchmark ecall number that results in exits.
+global_asm!(
+    r#"
+.text
+.align 4
+.global _start
+_start:
+    csrw mie, 0x1      // Dummy instruction to exit firmware
+    li a6, 3           // Miralis ABI FID: benchmark
+    li a7, 0x08475bcd  // Miralis ABI EID
+    ecall
+"#,
+);
+
+firmware_panic!();
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -216,7 +216,6 @@ impl Benchmark {
     }
 
     /// Log counter value. Tag allows to link different recordings together for analysis.
-    #[allow(dead_code)]
     pub fn record_counter(counter: Counter, tag: &str) {
         if !config::BENCHMARK {
             return;
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -3,7 +3,7 @@
 //! This is useful for creating different benchmark on time of execution or
 //! the number of instruction for example.
 
-use core::arch::asm;
+use spin::Mutex;
 
 use crate::arch::{Arch, Architecture, Csr};
 use crate::config;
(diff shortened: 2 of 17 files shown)


src/main.rs
```diff
@@ -148,28 +151,36 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
...
 
             if config::BENCHMARK {
-                if config::BENCHMARK_NB_EXITS {
-                    Benchmark::record_entry("nb_exits", "nb_exits", ctx.nb_exits);
+                let mut bench = BENCH.lock();
+                bench.stop_interval_counter(IntervalCounter::ExecutionTime, "main::handle_trap");
+                bench.stop_interval_counter(IntervalCounter::InstructionRet, "main::handle_trap");
+                bench.increment_counter(Counter::TotalExits);
+
+                // TODO : remove this and call from an ecall
```
CharlyCst, line 178: Just a note, no need to spend time fixing it:
In English there is no space before ":" like there is in French, so it would be `TODO:`. Some editors actually highlight comments of the form `XXX:` but not `XXX :`, so it's a good practice to follow the convention.

src/main.rs
```diff
@@ -148,28 +151,36 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
...
 fn main_loop(ctx: &mut VirtContext, mctx: &mut MiralisContext) -> ! {
-    let mut bench = Benchmark::new();
-
     loop {
         unsafe {
             if config::BENCHMARK {
-                bench.start();
+                let mut bench = BENCH.lock();
+                bench.start_interval_counter(IntervalCounter::ExecutionTime);
+                bench.start_interval_counter(IntervalCounter::InstructionRet);
```
CharlyCst, line 159: Do we want to manually call the start and stop multiple time or could the function start/stop all enabled counters at once?

src/main.rs
```diff
@@ -44,6 +45,8 @@ extern "C" {
 #[allow(non_upper_case_globals)]
 pub(crate) static _stack_start: u8 = 0;
 
+static BENCH: Mutex<Benchmark> = Mutex::new(Benchmark::new());
```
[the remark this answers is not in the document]
CharlyCst: Another important comment, but we can handle that later when we move to multi-core: this would need to be thread local, not a shared global. Otherwise multiple cores would increment the same counters, which might not be what we want.

Setting-up thread local variables is a bit complicated though, so we can delay that for now until we need to benchmark on multicore machines.

src/main.rs
```diff
@@ -148,28 +151,36 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
...
             if config::BENCHMARK {
-                if config::BENCHMARK_NB_EXITS {
-                    Benchmark::record_entry("nb_exits", "nb_exits", ctx.nb_exits);
+                let mut bench = BENCH.lock();
+                bench.stop_interval_counter(IntervalCounter::ExecutionTime, "main::handle_trap");
+                bench.stop_interval_counter(IntervalCounter::InstructionRet, "main::handle_trap");
+                bench.increment_counter(Counter::TotalExits);
+
+                // TODO : remove this and call from an ecall
+                if ctx.nb_exits % 50 == 0 {
```
CharlyCst, line 179: The number of exits before printing should probably be something that can be configured, just in case that value does not fit all use-cases.

src/main.rs
CharlyCst, line 1: See the individual comments, but you probably want to abstract the whole benchmarking a bit more. The whole locking and selecting individual counters and the checks for `BENCHMARK` enabled should probably be hidden within the `benchmark.rs` implementation.

CharlyCst (changes requested):
Thanks a lot for all the new changes :)

It's a big features so I have a few comments about design and implementation, but it's going in the right direction!

firmware/ecall_benchmark/main.rs
```diff
@@ -0,0 +1,22 @@
...
+use miralis_abi::firmware_panic;
+
+// This ecall Miralis with benchmark ecall number that results in exits.
+global_asm!(
+    r#"
+.text
+.align 4
+.global _start
+_start:
+    csrw mie, 0x1      // Dummy instruction to exit firmware
```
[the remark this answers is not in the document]
NoeTerrier: It's not requiered, but it result in a firmware exit: it's to show that the exit counter works :)

Maybe I can add an assertion to really check that the counter is set to 1 or remove this if it's useless here.

src/main.rs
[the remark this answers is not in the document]
NoeTerrier: Yes I was thinking about it and had the same conclusion.

src/main.rs
```diff
@@ -80,7 +83,7 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
     let hw = unsafe { Arch::detect_hardware() };
 
     // Initialize Miralis's own context
-    let mut mctx = MiralisContext::new(hw);
+    let mut mctx: MiralisContext = MiralisContext::new(hw);
```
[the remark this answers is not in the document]
NoeTerrier: We do not, my bad I think I was testing things on that line and didn't properly undo my changes :)

src/main.rs
NoeTerrier: It can be all at once sure, I'll remove the all or nothing.

src/main.rs
NoeTerrier: I removed the block as counters will probably logged at the end of the execution by ecall to Miralis.

src/benchmark.rs
```diff
@@ -22,73 +22,228 @@ macro_rules! benchmark_print {
...
-            let stop_time: usize;
-            unsafe {
-                asm!(
-                    "rdtime {value}", 
-                    value = out(reg) stop_time);
+        for counter in [
+            IntervalCounter::ExecutionTime,
+            IntervalCounter::InstructionRet,
+        ]
+        .map(Either::IntervalCounter)
```
CharlyCst, line 161: Wow I didn't know we could do that in Rust, thanks for the trick :)
That implies that `Either::IntervalCounter` is a function, which is cool.

CharlyCst (approved):
Thanks for the changes, I'll merge that tomorrow :)

Next step is going to get numbers for the paper with our new tools 🥳

PR #132, opened 2024-08-09
Add linux driver release

NoeTerrier:
Rework benchmark analysis code to handle multiple runs of a given binary.
Justfile command `just benchmark` can now receive an argument to control the number of run we want to execute.

### DIFF
--- a/misc/artifacts.toml
+++ b/misc/artifacts.toml
@@ -2,6 +2,7 @@
 
 [bin]
 opensbi = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.1.7/opensbi.bin"
-linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.0/opensbi-linux-kernel-exit.bin"
-linux-shell = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.0/opensbi-linux-kernel-shell.bin"
+linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-exit.bin"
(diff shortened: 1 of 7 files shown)


misc/artifacts.toml
```diff
@@ -2,6 +2,7 @@
 
 [bin]
 opensbi = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.1.7/opensbi.bin"
-linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.0/opensbi-linux-kernel-exit.bin"
-linux-shell = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.0/opensbi-linux-kernel-shell.bin"
+linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-exit.bin"
+linux-shell = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-shell.bin"
+linux-driver = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-driver.bin"
```
CharlyCst, line 7: Maybe a more explicit name would be helpful, by running `just run linux-driver` it is not clear what is supposed to happen.
Something like `linux-boot-benchmark` would be clearer I think.

On a somewhat related note, I am thinking that it would be helpful to have some documentation on the artifacts binary now that we have a few of them. I opened a low-priority issue for that: https://github.com/CharlyCst/miralis/issues/133

CharlyCst (changes requested):
Looks good to me, I just have one minor request which is to make the name of the new binary more explicit (see comment).

Thanks for the changes!

PR #134, opened 2024-08-12
Add csr write frimware for benchmarking

NoeTerrier:
Add a simple firmware that only performs writes to a csr in order to benchmark emulation handling of a csr write. Benchmarks aims to be tested on the board.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,6 +128,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "csr_write_benchmark"
+version = "0.1.0"
+dependencies = [
+ "miralis_abi",
+]
+
+[[package]]
 name = "default"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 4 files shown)


CharlyCst (approved):
Thanks!

Let's merge that one, but I think we will need to spend some time thinking about how we want to organize our benchmarks. One things we will want for sure is to be able to parametrize the number of runs using the config, and we will also want to run them in the CI with a small parameter (e.g. 10 runs instead of 10 000).
We will also want to put all the benchmarks in a single place, maybe in a `benchmark` folder within `./firmware`.

Another small thing (I'll handle it but just to let you know): when you make a mistake in you last commit you can fix it with `git commit --amend`. That will update your last commit (it will add the new changes there instead of creating a new one), so you will need to force push afterward.

NoeTerrier:
> Another small thing (I'll handle it but just to let you know): when you make a mistake in you last commit you can fix it with `git commit --amend`. That will update your last commit (it will add the new changes there instead of creating a new one), so you will need to force push afterward.

Thanks for the tip :)

PR #138, opened 2024-08-13
Document artifact binaries

### LINKED ISSUE #133: Document artifact binaries
As we build more and more artifacts over time (which is a good thing!) it starts getting harder to track which artifact is doing what, and from where it is built.

For now we use a simply toml file to store artifacts, like this:

```toml
[bin]
opensbi = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.1.7/opensbi.bin"
linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-exit.bin"
```

But it doesn't give any context.

A first step to improve the situation would be to comment each artifact, like this:

```toml
[bin]
# An OpenSBI image with a dummy payload that immediately exits.
opensbi = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.1.7/opensbi.bin"
# An OpenSBI image with a Linux payload that exits immediately after boot.
linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-exit.bin"
```

Something even better would be to make the documentation (and potentially other medatada) machine-readable by changing our artifact.toml format, for instance:

```toml
[bin.opensbi]
description = "An OpenSBI image with a dummy payload that immediately exits."
url = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.1.7/opensbi.bin"
repo = "https://github.com/CharlyCst/miralis-artifact-opensbi"

[bin.linux]
# ....
```

In the future that would open the door for potentially better user experience, such as printing a description when downloading/running an artifact, or generating documentation pages from the `artifact.toml` page.

NoeTerrier:
Add documentation and change toml as sugested in issue #133.

Close #133

### DIFF
--- a/misc/artifacts.toml
+++ b/misc/artifacts.toml
@@ -1,8 +1,26 @@
 # Artifact manifest
 
-[bin]
-opensbi = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.1.7/opensbi.bin"
-linux = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-exit.bin"
-linux-shell = "https://github.com/CharlyCst/miralis-artifact-opensbi/releases/download/v0.2.1/opensbi-linux-kernel-shell.bin"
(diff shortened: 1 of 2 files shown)


CharlyCst (approved):
Thank for the changes!

PR #139, opened 2024-08-13
Improve benchmark firmware organization

NoeTerrier:
Add a way to parametrize benchmark. The helpers for reading environment values have been factorized into a crate and can be now be called inside firmware.

Move benchmark firmware into firmware/benchmark/

Add write_csr benchmark into the CI.


config/qemu-virt-benchmark.toml
```diff
@@ -19,3 +19,4 @@ enable = true
 time = true
 instruction = true
 nb_exits = true
+nb_iter=10
```
NoeTerrier, line 22:
```suggestion
nb_iter = 10
```

firmware/benchmark/csr_write/main.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use core::arch::asm;
+
+use config_helpers::parse_usize_or;
+use miralis_abi::{setup_firmware, success};
+
+setup_firmware!(main);
+
+fn main() -> ! {
+    let iterations = parse_usize_or(option_env!("MIRALIS_BENCHMARK_NB_ITER"), 1);
```
CharlyCst, line 12: We will need to read this environment variable from multiple firmware images in the future, so we should probably factor it out in a single place. The `abi` crate is made for that, it defines functions for the firmware to interact with Miralis and other common utilities (such as initializing the stack, etc...)

runner/src/artifacts.rs
```diff
@@ -124,7 +124,37 @@ pub fn locate_artifact(name: &str) -> Option<Artifact> {
...
+    if let Some(artifact) = external.get(name) {
+        return Some(artifact.clone());
+    }
+
+    // Could not find artifact
+    None
+}
+
+/// Check if one entry match the name
+fn find_artifact(firmware_path: &PathBuf, name: &str) -> Option<Artifact> {
```
CharlyCst, line 157: Let's call it `find_artifact_source`. There are multiple kinds of artifacts (sources or binaries), here we only look for sources that we can compile.

CharlyCst (changes requested):
I have just some small remarks but looks good otherwise, thanks!

CharlyCst (approved):
Thanks for the changes, look all good :)

PR #140, opened 2024-08-13
Benchmark compute basic stats during runtime

NoeTerrier:
Modification of benchmark to allow stats to be computed during runtime to avoid periodic printing. This only allow firmware or payload that ends with a benchmark ecall to miralis to be benchmarked.

### DIFF
--- a/crates/abi/src/lib.rs
+++ b/crates/abi/src/lib.rs
@@ -39,6 +39,16 @@ pub fn failure() -> ! {
     }
 }
 
+/// Ask Miralis to end benchmark and print results.
+pub fn miralis_end_benchmark() -> ! {
+    unsafe { miralis_ecall(abi::MIRALIS_BENCHMARK_FID).ok() };
+
+    // Loop forever, this should never happen as Miralis will terminate the execution before.
+    loop {
+        hint::spin_loop();
+    }
+}
+
 // ————————————————————————————— Firmware Setup ————————————————————————————— //
 
 /// Configure the firmware entry point and panic handler.
--- a/docs/tooling.md
+++ b/docs/tooling.md
@@ -32,11 +32,13 @@ We do have an artifact named `opensbi`, so the runner will check if it is alread
 ## Benchmark
 
 Different benchmark can be run with `just benchmark`.
-This command computes and prints some statistics about a run of Miralis.
+This command computes and prints some statistics about a run of Miralis with a given binary.
 
-The default behaviour is counting the number of retired instructions of function `run_vcpu` and `handle_trap` called in `main.rs` with the `default` firmware. You can specify the binary you want to benchmark with `just benchmark <binaries>`. For example, if you want to benchmark `linux` you can run `just benchmark linux`. 
+The binary should call Miralis to print the benchmark results at the end of its execution. Statistics are computed during the runtime in a streaming maner.
 
-One can choose which benchmark to run in the `benchmark` section of the `config.toml` file. The field `enable` is overwritten to true when running this command.
+You can specify the binary you want to benchmark with `just benchmark <binaries>` (by default it is `ecall_benchmark`). For example, if you want to benchmark `linux` you can run `just benchmark linux`. 
 
-If you collect output of a run into a file, you can feed the file to the just `analyze-benchmark` command to get the statistics of the run.
+One can choose what to benchmark in the `benchmark` section of the `config.toml` file. The field `enable` is overwritten to true when running this command.
+
+If you collect output of a run into file (should be in csv format using `csv_format` in the config), you can feed the file to the just `analyze-benchmark` command to get the statistics of the run. You can also put multiple files of multiple runs into a folder and give the path of the folder. This will compute the average of all runs.
 
--- a/docs/tooling.md
+++ b/docs/tooling.md
@@ -31,14 +31,9 @@ We do have an artifact named `opensbi`, so the runner will check if it is alread
 
 ## Benchmark
 
-Different benchmark can be run with `just benchmark`.
-This command computes and prints some statistics about a run of Miralis with a given binary.
+One can choose to activate benchmarking and what to benchmark in the `benchmark` section of the `config.toml` file.
 
-The binary should call Miralis to print the benchmark results at the end of its execution. Statistics are computed during the runtime in a streaming maner.
+Statistics are computed during the runtime in a streaming maner, then printed at the end of the run, either in a fancy way or in csv format. The firmware should ecall Miralis with FID 3 in order to ends the benchmark before exiting.
 
-You can specify the binary you want to benchmark with `just benchmark <binaries>` (by default it is `ecall_benchmark`). For example, if you want to benchmark `linux` you can run `just benchmark linux`. 
-
-One can choose what to benchmark in the `benchmark` section of the `config.toml` file. The field `enable` is overwritten to true when running this command.
-
-If you collect output of a run into file (should be in csv format using `csv_format` in the config), you can feed the file to the just `analyze-benchmark` command to get the statistics of the run. You can also put multiple files of multiple runs into a folder and give the path of the folder. This will compute the average of all runs.
+If you collect the csv output of a run into file (should be in csv format using `csv_format` in the config), you can feed the file to the just `analyze-benchmark` command to get the statistics of the run. You can also put multiple files of multiple runs into a folder and give the path of the folder. This will compute the average of all runs.
 
--- a/justfile
+++ b/justfile
@@ -50,8 +50,8 @@ test:
 	cargo run --package runner -- run --config {{qemu_virt}} --firmware zephyr --max-exits 1000000
 
 	# Test benchmark code
-	cargo run --package runner -- run --config {{qemu_virt_benchmark}} --firmware csr_write --benchmark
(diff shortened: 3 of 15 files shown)


docs/tooling.md
```diff
@@ -32,11 +32,13 @@ We do have an artifact named `opensbi`, so the runner will check if it is alread
...
 
-The default behaviour is counting the number of retired instructions of function `run_vcpu` and `handle_trap` called in `main.rs` with the `default` firmware. You can specify the binary you want to benchmark with `just benchmark <binaries>`. For example, if you want to benchmark `linux` you can run `just benchmark linux`. 
+The binary should call Miralis to print the benchmark results at the end of its execution. Statistics are computed during the runtime in a stream maner.
 
-One can choose which benchmark to run in the `benchmark` section of the `config.toml` file. The field `enable` is overwritten to true when running this command.
+You can specify the binary you want to benchmark with `just benchmark <binaries>` (by default it is `ecall_benchmark`). For example, if you want to benchmark `linux` you can run `just benchmark linux`. 
 
-If you collect output of a run into a file, you can feed the file to the just `analyze-benchmark` command to get the statistics of the run.
+One can choose what to benchmark in the `benchmark` section of the `config.toml` file. The field `enable` is overwritten to true when running this command.
```
CharlyCst, line 41: It is not super clear what `enable` does just from reading the documentation.
You should probably invert the text, saying that benchmarks can be enables with `benchmark.enable`, that different values can be measures, and finally that `just benchmark` overrides `benchmark.enable` to `true`so it is equivalent to `just run` with a configuration that enables benchmark.

I have a related question: is there any difference now between using `just benchmark` and `just run` with a config that enables benchmarks?
I am pondering if we should stick with `just benchmark` or rather expose a configuration with benchmark enabled. Benchmarks only make sense on a board, but `just benchmark` is for QEMU, so I am just trying to evaluate if it is worth to keep it.

firmware/csr_write_benchmark/main.rs
```diff
@@ -15,5 +15,8 @@ fn main() -> ! {
             );
         }
     }
+    unsafe {
+        let _ = miralis_ecall(3);
```
CharlyCst, line 19: See other comment in `abi` crate. We should rather expose a `miralis_print_benchmark` function. The function can do things such as panic on failure as no failure is expected with this ecall.

justfile
```diff
@@ -85,8 +85,12 @@ install-toolchain:
 	rustup component add llvm-tools-preview --toolchain "$(cat rust-toolchain)"
 	cargo install cargo-binutils
 
+# The following line gives highlighting on vim
+# vim: set ft=make :
```
CharlyCst, line 89: Keep that one at the end, it gives automatic highlighting in vim (and some other editors) by configuring the file-type to makefile (which is sort of similar enough to justfiles).

I added it back on main as it got deleted on another branch :)

justfile
```diff
@@ -85,8 +85,12 @@ install-toolchain:
...
+
+# Run a firwmare with benchmark settings. The firmware should call for benchmark printing before exiting
 benchmark firmware=benchmark iterations=default_iterations:
-	cargo run --package runner -- run -v --firmware {{firmware}} --benchmark --benchmark-iterations {{iterations}}
+	cargo run --package runner -- run -v --config {{qemu_virt_benchmark}} --firmware {{firmware}} --benchmark --benchmark-iterations {{iterations}}
```
CharlyCst, line 93: Is `--benchmark-iterations` overwriting `nb_iter` in the config?

I am just thinking that we should probably keep the same names for consistency, but we can do that in a separate commit/PR.

src/benchmark.rs
```diff
@@ -2,6 +2,7 @@
 //!
 //! This is useful for creating different benchmark on time of execution or
 //! the number of instruction for example.
+use core::usize;
```
CharlyCst, line 5: Not needed, I know sometimes Rust-Analyzer adds those without asking ^^'

src/benchmark.rs
```diff
@@ -161,89 +196,136 @@ impl Benchmark {
...
+        counter: &IntervalCounter,
+        scope: &Scope,
+        value: usize,
+    ) {
+        let index = Self::interval_counter_index(counter, scope);
+        let stats = &mut self.interval_counters[index];
+        stats.count += 1;
+        stats.sum += value;
+        stats.mean = stats.sum / stats.count;
+        stats.min = if value < stats.min { value } else { stats.min };
```
CharlyCst, line 241: You can replace those by [`core::cmp::min`](https://doc.rust-lang.org/nightly/core/cmp/fn.min.html), and manual max by `core::cmp::max`.

On some architecture it is possible to have a better implementation than using an `if` (using branch-less instructions such as conditional moves which do not pollute the CPU branch predictor). So using the built-in Rust function gives us the best implementation we can have.

CharlyCst (commented):
Thanks for the changes, I have added a few comments and questions, but things are going in the right direction!

docs/tooling.md
NoeTerrier: I see, the command is still useful because it calls the runner with --benchmark arguments and a number of runs it has to do, in order to have an average of all runs. It is still only on qemu tho. If we don't need to benchmark on qemu then it is useless.

justfile
NoeTerrier: No, it's two different things:

- `--benchmark-iterations` is the times the runner runs the entire triplet (miralis, payload and firmware) to compute average on the entire thing at the ends. It is only usefull for qemu benchmarking.
- `nb_iter` is an argument that is passed to our custom firmwares (that benchmarks only one thing many times). It can be used differently between firmwares (e.g. run `nb_iter` instructions, or `nb_iters` times a complex function or ecall). It is also used for building these firmwares in order to run on the board.

We can remove the first one and the just benchmark command if we don't want to benchmark (or not on many iterations) on qemu anymore.

src/benchmark.rs
NoeTerrier: Oh that's a strange behavior from Rust-Analyzer haha

src/benchmark.rs
NoeTerrier: Thank you! :)

crates/abi/src/lib.rs
```diff
@@ -120,21 +120,12 @@ macro_rules! firmware_panic {
...
-        out("a1") value,
     );
 
-    if error != 0 {
-        Err(error)
-    } else {
-        Ok(value)
-    }
+    panic!("Should not return from ending benchmarking.")
 }
```
CharlyCst, line 131: Sorry I might not have been super clear, we need to keep the `miralis_ecall` as it is used by other functions, but not make it public. You can create a `miralis_end_benchmark` function (and make it public) that calls into `miralis_ecall` internally.

You can put it above in the file with the other public-facing functions:

https://github.com/CharlyCst/miralis/blob/fef8730a352c367af5897810caa0c44136239570/crates/abi/src/lib.rs#L19-L40

justfile
CharlyCst: I see, that make sense thanks :)

Yes I am wondering if we want to keep the benchmarking capabilities for QEMU, we saw from experience with the board that the QEMU numbers don't make sense. Running the benchmarks on QEMU is still useful to test the code path and ensure nothing is broken, but it's just that the benchmark results don't make any sense so I am not sure if it makes sense to keep `--benchmark-iterations` for instance considering that we can't do anything with numbers from QEMU (contrary to what I hopped initially...). What do you think? Anyway, if we change something it can be done in a different PR.

CharlyCst (commented):
Thanks for the changes, just a small details and things look good :)

There has been some changes on `main` (I merged a bunch of PRs since), so if possible can you give it a try at rebasing? `git rebase main` should do the trick but let me know if things gets out of hand.

justfile
NoeTerrier: If the number doesn't make any sens, it is better to remove ´--benchmark-iterations´, and the ´just benchmark´ command, and only keep the ´--benchmark´ argument for the runner in the tests to test benchmark code. It will be simpler and more understandable then.

crates/abi/src/lib.rs
NoeTerrier: My bad, you were clear enough, I just forgot that.

crates/abi/src/lib.rs
```diff
@@ -39,6 +39,17 @@ pub fn failure() -> ! {
     }
 }
 
+#[inline]
+pub unsafe fn miralis_end_benchmark() -> ! {
```
CharlyCst, line 43: A few last details on this function:
- You can make it safe, it can't cause undefined behavior or doesn't have any pre-conditions that needs to be enforced by the caller.
- Add a small doc comment too, it's a public function in a library so it's always a good practice to have one.
- Instead of having inline assembly you can use the `miralis_ecall` helper, like `success` and `failure` do (`unsafe { miralis_ecall(abi::MIRALIS_FAILURE_FID).ok() };` but with the right FID constant)

After that I think we are good :)

CharlyCst (commented):
Looks good to me, let me know if you manage to run it successfully on the board and we can proceed to merge the changes :)

CharlyCst (approved):
Thanks for the changes :)

I'll merge the PR when I'll get some time.

Issue #135, opened 2024-08-12
Extend config to enable selecting the firmware start address


CharlyCst:
We already have a way to select Miralis start address in the config, but the firmware start address is hardcoded.

The config already has a `platform.firmware_address` parameter that we could use to set the start address of the firmware. To do so we could take inspiration from the way we set the start address of Miralis:

First we make it a variable in the linker script, see https://github.com/CharlyCst/miralis/blob/a29da2013dfdc3478e4d114b25c047d432dc579d/misc/linker-script.x#L3-L5.

Then we add some arguments when we build the firmware to create a variable with the right value, like we do here when building Miralis:
https://github.com/CharlyCst/miralis/blob/a29da2013dfdc3478e4d114b25c047d432dc579d/runner/src/artifacts.rs#L180

CharlyCst:
Fixed in #141.

PR #141, opened 2024-08-13
Firmware setup for the board

Wesdcv:
Tweaked the firmware setup a bit for it to work on the board properly

- *bss sections of firmware are now zeroed-out as well as miralis' sections, fixed symbols' alignment 
- Removed duplicate assignment of firmware and miralis's start addresses (in linker script and in platform definition)
- Added possibility to only build certain firmware for certain configuration, without actually running it as a payload in qemu (useful feature to build firmware for the board). Refs #135?

### DIFF
--- a/justfile
+++ b/justfile
@@ -50,6 +50,9 @@ test:
 	# Test benchmark code
 	cargo run --package runner -- run --config {{qemu_virt_benchmark}} --firmware ecall_benchmark --benchmark
 
+	# Test firmware build
+	just build-firmware {{qemu_virt}} default
+
 # Run unit tests
 unit-test:
 	cargo test --features userspace -p miralis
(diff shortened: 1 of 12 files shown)


justfile
```diff
@@ -62,6 +62,9 @@ run firmware=default:
 build config:
 	cargo run --package runner -- build --config {{config}}
 
+build-firmware config firmware=default:
```
CharlyCst, line 65: In a `justfile` it is important to add a comment on top of each top-level rule: it is display as documentation when running `just --list` (which is our default command). You can try to run `just` without any argument to see the current list of commands.

I realize that we are missing documentation on some other commands, so I'll add them too in the future :)

runner/src/main.rs
```diff
@@ -62,6 +62,8 @@ struct BuildArgs {
     #[arg(long)]
     /// Path to the configuration file to use
     config: Option<PathBuf>,
+    #[arg(short, long, default_value = "none")]
+    firmware: String,
```
CharlyCst, line 66: Here you can use `Option<String>` instead of relying on a default value of `"none"`.

CharlyCst (approved):
Thanks for the changes :)

I only have two minor comments, I'll handle them and merge the PR.

PR #142, opened 2024-08-14
Fix SEIE bit overwrite

### LINKED ISSUE #130: Linux loop issue with sifive-u54 cpu after booting.
Using [sifive-u54](https://static.dev.sifive.com/U54-MC-RVCoreIP.pdf) cpu as the virtual cpu for qemu, linux boots but randomly falls in what appears to be an infinite loop. 

To reproduce: checkout commit `b7f014fe04fd78294af1ad965ec3a2a7b2dc9fad`, add `.arg("-cpu").arg("sifive-u54")` to the runner's QEMU arguments and run `just run linux`. (with info log level and sufficient number of exits)

**Result**: after few tries, it eventually ends up in a loop and finishes by maximal number of exits:

```
...
[    0.355441] 9pnet: Installing 9P2000 support
[    0.355751] Key type dns_resolver registered
[    0.382373] debug_vm_pgtable: [debug_vm_pgtable         ]: Validating architecture page table helpers
[    0.389111] clk: Disabling unused clocks
[    0.389492] PM: genpd: Disabling unused power domains
[    0.389709] ALSA device list:
[    0.389829]   No soundcards found.
[    0.431572] Freeing unused kernel image (initmem) memory: 3032K
[    0.432497] Run /init as init process

Hello from Linux!

[Error | miralis] Reached maximum number of exits: 99999
error: Recipe `run` failed on line 48 with exit code 1
```

**Expected**: Linux boots and power down gracefully:
```
...
[    0.328417] 9pnet: Installing 9P2000 support
[    0.328751] Key type dns_resolver registered
[    0.352291] debug_vm_pgtable: [debug_vm_pgtable         ]: Validating architecture page table helpers
[    0.358122] clk: Disabling unused clocks
[    0.358497] PM: genpd: Disabling unused power domains
[    0.358718] ALSA device list:
[    0.358846]   No soundcards found.
[    0.395570] Freeing unused kernel image (initmem) memory: 3032K
[    0.396239] Run /init as init process

Hello from Linux!

[    0.450809] reboot: Power down
```

NoeTerrier:
SEIE bit of mip register was overwritten when context switches by a wrong value. This value comes from the fact that csrr instruction is modified for SEIE and output the logical OR of what is in mip and if there is a signal from an external device (here, it was the PLIC).

We then sync this bit between physical and virtual mip and take care to not take a wrong value of SEIE:

Close #130

### DIFF
--- a/src/virt.rs
+++ b/src/virt.rs
@@ -497,8 +497,12 @@ impl VirtContext {
         self.csr.mcause = self.trap_info.mcause;
         self.csr.mstatus = self.trap_info.mstatus;
         self.csr.mtval = self.trap_info.mtval;
-        self.csr.mip = self.trap_info.mip;
         self.csr.mepc = self.trap_info.mepc;
+
+        // Real mip.SEIE bit should not be different from virtual mip.SEIE as it is read-only in S-Mode or U-Mode.
+        // But csrr is modified for SEIE and return the logical-OR of SEIE and the interrupt signal from interrupt
+        // controller. (refer to documentation for further detail).
+        self.csr.mip = self.trap_info.mip & !mie::SEIE_FILTER | self.csr.mip & mie::SEIE_FILTER;
 
         match self.mode {
             Mode::M => {
@@ -751,7 +755,12 @@ impl VirtContext {
         Arch::write_csr(Csr::Medeleg, 0); // Do not delegate any exceptions
 
         self.csr.mie = Arch::read_csr(Csr::Mie);
-        self.csr.mip = Arch::read_csr(Csr::Mip);
+
+        // Real mip.SEIE bit should not be different from virtual mip.SEIE as it is read-only in S-Mode or U-Mode.
+        // But csrr is modified for SEIE and return the logical-OR of SEIE and the interrupt signal from interrupt
+        // controller. (refer to documentation for further detail).
+        self.csr.mip =
+            Arch::read_csr(Csr::Mip) & (!mie::SEIE_FILTER) | self.csr.mip & mie::SEIE_FILTER;
 
         // Remove Firmware PMP from the hardware
         mctx.pmp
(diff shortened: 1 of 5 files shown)


src/virt.rs
```diff
@@ -497,7 +497,7 @@ impl VirtContext {
         self.csr.mcause = self.trap_info.mcause;
         self.csr.mstatus = self.trap_info.mstatus;
         self.csr.mtval = self.trap_info.mtval;
-        self.csr.mip = self.trap_info.mip;
+        self.csr.mip = self.trap_info.mip & !mie::SEIE_FILTER | self.csr.mip & mie::SEIE_FILTER; // do not take SEIE into account
```
CharlyCst, line 500: Could you add a bit more details about why we do not take it into account?

You can mode it below `self.csr.mepc = ...` to make some space for a multi-line comment if needed. But this is a tricky part, so it needs to be commented with enough detail for someone coming later to understand what is going on.
In particular, it is clear from the code **what** you are doing here, but it is not clear **why** you are doing it. Adding a comment talking briefly about the two SEIE bits and which one we want to keep would help.

src/virt.rs
```diff
@@ -751,7 +751,8 @@ impl VirtContext {
         Arch::write_csr(Csr::Medeleg, 0); // Do not delegate any exceptions
 
         self.csr.mie = Arch::read_csr(Csr::Mie);
-        self.csr.mip = Arch::read_csr(Csr::Mip);
+        self.csr.mip =
```
CharlyCst, line 754: Same remark here, it is not clear **why** it can be wrong. Someone new on the project can't deduce it from the comment (it is clearly a non-obvious part of the RISC-V spec).
Adding a bit more explanation about the why would help.

src/virt.rs
```diff
@@ -1001,7 +1002,22 @@ impl HwRegisterContextSetter<Csr> for VirtContext {
...
+                            Arch::clear_csr_bits(Csr::Mip, mie::SEIE_FILTER);
+                        }
+                    } else {
+                        unsafe {
+                            Arch::set_csr_bits(Csr::Mip, mie::SEIE_FILTER);
+                        }
+                    }
+                }
+                self.csr.mip = value;
+            }
```
CharlyCst, line 1020: We also need some good commenting here, why are we doing that? Explain a bit that there are two SEIE bits in different places and that we need to keep the hardware in sync and how you do it.

CharlyCst (changes requested):
Thanks a lot for the patch, it was definitely a tricky one!

I thing we need more documentation about how the SEIE bit works and how we do things to work around it. It is a non-trivial issue so we need to keep the knowledge around in case we run into more issue in the future.
I added a few comments to request more inline comments, but I think we will need to add a section in the documentation to explain the problem and our solution too (if possible listing the other solutions we discussed in case we need to re-visit them in the future). You can add a new subsection in `docs/architecture.md`, it seems like a good place.

CharlyCst (approved):
Thanks, looks all good to me :)

I'll proceed to the merge.

PR #143, opened 2024-08-14
[WIP] Build miralis as a firmware

francois141:


justfile
```diff
@@ -69,6 +69,9 @@ build config:
 build-firmware config firmware:
 	cargo run --package runner -- build -v --config {{config}} --firmware {{firmware}}
 
+miralis-firmware config:
+	cargo run --package runner -- build --config {{config}} --as-firmware true
+
```
CharlyCst, line 74: Do we need a new `just` rule or can we use the already existing `just build-firmware` with `miralis` as argument? It seems you changed the runner to recognize `miralis` as a special case, so we might not need the extra `just` rule :)

runner/src/artifacts.rs
```diff
@@ -211,7 +227,7 @@ fn objcopy(target: &Target) -> PathBuf {
     let mut bin_path = path.clone();
 
     match target {
-        Target::Miralis => {
+        Target::Miralis | Target::MiralisAsFirmware => {
             elf_path.push("miralis");
             bin_path.push("miralis.img");
```
CharlyCst, line 232: Doesn't `miralis.img` get overwritten when we build Miralis twice? In the end we need two images, one as the host Miralis and one as the virtualized Miralis, how do we ensure we keep both?

CharlyCst: Looking at the code below it seems they use two different targets, so they end up in two different directories which solves the issue :)

runner/src/build.rs
```diff
@@ -12,13 +12,21 @@ pub fn build(args: &BuildArgs) {
...
         println!("Built firmware, binary available at:");
         println!("{}", firmware.display());
     } else {
-        let miralis = build_target(Target::Miralis, &cfg);
+        let mut target = Target::Miralis;
+        if let Some(as_firmware) = &args.as_firmware {
+            if *as_firmware {
+                target = Target::MiralisAsFirmware;
+            }
+        }
```
CharlyCst, line 27: Do we need that special case if we can build Miralis by passing `miralis` as the firmware name?

runner/src/main.rs
```diff
@@ -65,6 +65,9 @@ struct BuildArgs {
     /// Build a firmware instead of Miralis
     #[arg(short, long)]
     firmware: Option<String>,
+    #[arg(long)]
+    /// Build miralis as firmware
+    as_firmware: Option<bool>,
```
CharlyCst, line 70: We might not need this extra flag if `miralis` can be passed as the firmware name.

CharlyCst (commented):
Thanks for the changes!

Before merging I have two small questions:
- I assume for now the virtualized Miralis does not itself run a virtualized firmware (i.e. nested virtualization)?
- Ca we run Miralis as a firmware in the CI? For that we would need a way to know if the execution was successful. We have a small crate (`abi`) that we use for communicating between a firmware and Miralis, we could use it to exit with a success or an error.

I am thinking that we might need to create a new platform (something like `miralis_virt`) to use when running Miralis as a firmware. That way we could have a specialized implementation of logging (Miralis exposes logging capabilities to firmware) and to properly exit with a success/failure code. But that is something we can think about as a second step.

=========== [Edit: @francois141 ] ===========

You are welcome! 
- Indeed, virtualized Miralis doesn't run a firmware at the moment. I think this could be the second step.
- Done, however compiling the crate('abi') with Miralis seems to introduced a few painful issues with the unit tests. In the meantime, I have a small temporary workaround. In case you know how to fix it, feel free to let me know.

```
error: invalid register a0: unknown register
  --> crates/abi/src/logger.rs:47:21
```

Regarding the creation of a new platform for specialized logging. I am a bit skeptical because a specialized "firmware" logger should be present not only in qemu but also on the board. We don't want to have 2 platforms per emulator. Instead I would suggest to have two logging implementation per use case (emulator vs firmware).  Also I think if the goal is to run Miralis on top of Miralis, I feel it makes more sense to keep the original logger rather than having a system call as the goal it to run Miralis on top of Miralis  :) Do you see what I mean?

justfile
francois141: You are right about that. I delete this extra command.

runner/src/artifacts.rs
francois141: Exactly, we don't have any issues even if the name is the same

runner/src/main.rs
francois141: Agree, se previous comment

runner/src/build.rs
francois141: You are right. I had a look at my code again and I realized it can be a bit simplified.

francois141:
@CharlyCst While working on that feature, I found an issue with mtime register: https://github.com/CharlyCst/miralis/issues/154. We need to fix that before being able to make progress on that feature.

CharlyCst:
Regarding

> Done, however compiling the crate('abi') with Miralis seems to introduced a few painful issues with the unit tests. In the meantime, I have a small temporary workaround. In case you know how to fix it, feel free to let me know.

This is because `abi` is using RISC-V assembly. Miralis can be built for other platforms than RISC-V (we use that for running unit tests for instance, by compiling Miralis to the host architecture).
What we do to make this work is that we regroup all the assembly in the `arch` folder and we maintain two implementation: one for RISC-V and one for other architecture (which we call 'userspace').

I realize that this might not be a practical solution when including a library.
Another option is to make the library compile on other architecture, which seems more practical to me (there are only 2 assembly snipets that we can hide behind `#[cfg]`).

For instance we can panic when trying to do an `ecall` on another platform than RISC-V:

```rs
    // On non RISC-V platform we panic here
    #[cfg(not(target_arch = "riscv64"))]
    panic!("Tried to use `ecall` on non RISC-V archiecture");

    // On RISC-V platforms we perform the ecall normally 
    #[cfg(target_arch = "riscv64")]
    asm!(
        "ecall",
        in("a6") fid,
        in("a7") abi::MIRALIS_EID,
        out("a0") error,
        out("a1") value,
    );
```

The trick here is that `#[cfg(...)]` will do conditional compilation, like `#ifdef` in C. We try to avoid those as much as possible (because they are hard to debug and needs to be tested in all configurations), but when compiling with assembly snippets on different platform it's the only solution.

The other big assembly snippets in `abi` are hidden behind macros, so they won't cause issue when cross-compiling if we don't use the macros in Miralis.

runner/src/artifacts.rs
```diff
@@ -222,6 +250,10 @@ pub fn build_target(target: Target, cfg: &Config) -> PathBuf {
                     }
                 }
             }
+
+            if target == Target::MiralisAsFirmware {
+                build_cmd.arg("--features").arg("miralis_as_firmware");
```
CharlyCst, line 255: As mentioned above let's rename it `platform_miralis` :)

src/logger.rs
```diff
@@ -55,6 +55,22 @@ impl Logger {
...
+        };
+
+        Plat::debug_print(format_args!(
+            "[{} | {}{}] {}\n",
+            level_display(record.level()),
+            record.target(),
+            display_from_firmware,
+            record.args()
+        ));
+    }
```
CharlyCst, line 73: Let's create a platform for Miralis, that way we can just implement `Plat::debug_print` there.

src/logger.rs
```diff
@@ -69,12 +85,7 @@ impl log::Log for Logger {
 
         if global_level_mask || module_level_mask {
             // Writes the log
-            Plat::debug_print(format_args!(
-                "[{} | {}] {}\n",
-                level_display(record.level()),
-                record.target(),
-                record.args()
-            ))
+            Self::print_log(record);
```
CharlyCst, line 88: We don't need this anymore if we add a Miralis platform.

src/main.rs
```diff
@@ -155,6 +155,16 @@ pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> !
         ctx.pc = firmware_addr;
     }
 
+    // In case we compile Miralis as firmware, we stop execution at that point for the moment
+    // This allows us to run Miralis on top as an integration test for the moment
+    // In the future, we plan to run Miralis "as firmware" running a firmware
+    #[cfg(as_firmware)]
+    {
+        use miralis_abi::success;
+        success();
```
CharlyCst, line 164: Once we have a Miralis platform we can use `Plat::exit_success` :)

CharlyCst (commented):
Great work, thanks!

I think is taking a great direction :)
This is a big change which touches quite a lot of the project's architecture, so it's normal to go through a bunch of iterations to get things right, but I think we are getting close. Thanks for your patience!

There are only two main aspects that I think we can improve:
- Limiting the proliferation of `#[cfg]`s in the `abi` crate. We can limit them two simple functions: `miralis_ecall` and a new `miralis_ecall3` (see comments).
- Creating a Miralis platform (like we have a QEMU `virt` and a `visionfive2` platform). That would better fit our current architecture (which has been designed to introduce more platforms over time) and I actually think it makes sense (as we are indeed running on top of Miralis which exposes a different platform interface).

CharlyCst:
As a heads-up, we are going to merge another PR that changes quite a bit the architecture of the project (#145) soonish. As both PRs are getting ready around the same time there might be some conflicts (hopefully not too many).

When comes the time to merge I'll handle the rebase, as I have been following both PRs :)

CharlyCst (dismissed):
Looks great to me :)

Thanks for pushing that all the way to the end, it's a super cool feature! It's a good demo of what is possible with virtualization and it's going to be useful for testing.

I will handle the rebase and merge with the other PR when I get some time.

PR #144, opened 2024-08-15
Added option to build with a release profile (on by default)

Wesdcv:
- Release-mode build is a default mode for building miralis and firmware.
- You can enable debug mode (dev profile) by setting `debug_compile = true` in the corresponding config.toml file.
- Release profile builds in a `release` folder instead of `debug`.
- Release profile enables all avaliable runtime optimizations, which may increase build times.
- There was an issue with Mcause firmware failing some assert checks with `incremental = false` (default option for release), so for now we explicitly set `incremental = true`. It affects the size of the binary a bit (~2kb).
- In debug mode the resulting `miralis.img` file weights _76272_ byes, in release mode it's _58416_.
- There is a question to what extent should we optimize the release build (e.g. should we strip symbols from the executables?).

Execution times for default firmware on QEMU (just as a rough estimate) :

Release:
```
real    0m0.109s
user    0m0.070s
sys     0m0.037s
```
Debug:
```
real    0m0.121s
user    0m0.074s
sys     0m0.045s
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ opt-level = 3
 
 [profile.release]
 panic = "abort"
+codegen-units = 1
 
 [workspace]
 members = [
--- a/config/example.config.toml
+++ b/config/example.config.toml
@@ -29,8 +29,6 @@ color = true
 # Maximum number of firmware exits before terminating.
 # No maximum cap if not present
 max_firmware_exits = 400
-# Option to compile using release profile (dev profile is set by default)
-debug = false
 
 [vcpu]
 # Enable support for S_mode
@@ -45,14 +43,6 @@ max_pmp = 8
 # Name of the platform (i.e. board) to compile for.
 # Default to "qemu_virt"
 name = "qemu_virt"
-
-# Miralis binary will be compiled with this value as a start address
-# Default to "0x80000000"
-start_address = 0x80000000
-
-# Firmware binary will be compiled with this value as a start address
-# Default to "0x80200000"
-firmware_address = 0x80200000
 
 # Nuber of harts (i.e. cores).
 # Default to 1.
@@ -86,3 +76,19 @@ world_switches = false
 # Number of iterations to be used by benchmark firmware.
 # What is iterated on may vary from one firmware to another.
 nb_iter = 1000
+
+[target.miralis]
+# Build profile for Miralis (dev profile is set by default)
+profile = "dev"
+
+# Miralis binary will be compiled with this value as a start address
+# Default to "0x80000000"
+start_address = 0x80000000
+
+[target.firmware]
+# Build profile for the firmware (dev profile is set by default)
+profile = "dev"
+
+# Firmware binary will be compiled with this value as a start address
+# Default to "0x80200000"
+start_address = 0x80200000
--- a/config/example.config.toml
+++ b/config/example.config.toml
@@ -29,6 +29,8 @@ color = true
 # Maximum number of firmware exits before terminating.
 # No maximum cap if not present
 max_firmware_exits = 400
+# Option to compile using release profile (dev profile is set by default)
+debug = false
 
 [vcpu]
 # Enable support for S_mode
--- a/justfile
+++ b/justfile
@@ -1,4 +1,5 @@
 default          := "default"
+config           := "config.toml"
 benchmark        := "ecall_benchmark"
 qemu_virt        := "./config/test/qemu-virt.toml"
 qemu_virt_2harts := "./config/test/qemu-virt.toml"
@@ -51,22 +52,22 @@ test:
 	cargo run --package runner -- run --config {{qemu_virt_benchmark}} --firmware ecall_benchmark --benchmark
 
 	# Test firmware build
-	just build-firmware {{qemu_virt}} default
+	just build-firmware default {{qemu_virt}}
 
 # Run unit tests
 unit-test:
 	cargo test --features userspace -p miralis
 
 # Run Miralis
-run firmware=default:
-	cargo run --package runner -- run -v --firmware {{firmware}}
+run firmware=default config=config:
+	cargo run --package runner -- run -v --config {{config}} --firmware {{firmware}} 
 
 # Build Miralis with the provided config
 build config:
 	cargo run --package runner -- build --config {{config}}
 
 # Build a given firmware with the provided config
-build-firmware config firmware:
+build-firmware firmware config=config:
 	cargo run --package runner -- build -v --config {{config}} --firmware {{firmware}}
 
 # Run Miralis but wait for a debugger to connect
--- a/justfile
+++ b/justfile
@@ -4,6 +4,7 @@ benchmark        := "ecall_benchmark"
 qemu_virt        := "./config/test/qemu-virt.toml"
 qemu_virt_2harts := "./config/test/qemu-virt.toml"
 qemu_virt_benchmark := "./config/test/qemu-virt-benchmark.toml"
+qemu_virt_release := "./config/test/qemu-virt-release.toml"
 benchmark_folder := "./benchmark-out"
 default_iterations := "1"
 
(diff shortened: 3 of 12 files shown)


CharlyCst:
So it seems to be a bit faster indeed, good to know :)

Do you remember what was the issue with `incremental = false`?

CharlyCst:
Regarding the symbols, they are stripped from the executable by the next build step, which takes the ELF and create a binary image (by removing all the ELF metadata)

Cargo.toml
```diff
@@ -21,6 +21,16 @@ opt-level = 3
...
+opt-level = 3
+codegen-units = 1
+debug = 0
+debug-assertions = false
+# This is false by default for the release mode
+# However, false makes some assertions in Mcause firmware fail
+incremental = true
+lto = "fat"
+# Do we want to leave symbols?
+strip = "symbols"
```
CharlyCst, line 33: No need to strip symbols, they are already stripped when building the image (which is extracted from the ELF, without the metadata).

config/example.config.toml
```diff
@@ -29,6 +29,8 @@ color = true
 # Maximum number of firmware exits before terminating.
 # No maximum cap if not present
 max_firmware_exits = 400
+# Option to compile using dev profile (release profile is set by default)
+debug_compile = false
```
CharlyCst, line 33: Usually this is simply called `debug` in other projects, so I would tend to use the same name.

And sorry for changing my mind, but after reflection I think we should compile in debug by default, not release (see comment on the issue).

runner/src/config.rs
```diff
@@ -45,6 +45,7 @@ pub struct Log {
 #[serde(deny_unknown_fields)]
 pub struct Debug {
     pub max_firmware_exits: Option<usize>,
+    pub debug_compile: Option<bool>,
```
CharlyCst, line 48: I would tend to call that one simply `debug`, it is often how it is called in other projects.

runner/src/gdb.rs
```diff
@@ -26,7 +26,7 @@ static GDB_EXECUTABLES: &[&'static str] = &[
 /// need for such a function.
 fn build_gdb_command(gdb_executable: &str) -> Command {
     // Retrieve the path of Miralis's binary
-    let mut miralis_path = get_target_dir_path(&Target::Miralis);
+    let mut miralis_path = get_target_dir_path(&Target::Miralis, Mode::Debug); // Could there be any cases when we want to invoke gdb on a release mode build?
```
CharlyCst, line 29: Yes we might want to run GDB with a release builds, for instance to debug an issue that appears only when compiling in release mode (which can happen in projects with `unsafe` rust code when we do not properly enforce the safety requirements).

CharlyCst (commented):
Thanks for the changes, just a few comments :)

CharlyCst:
Ho strange, the error is in the firmware, not Miralis?

Looking at the test firmware it simply checks that we can read/write into `mcause`, that really should not be affected by incremental compilation or not.

I suspect the error comes from here: https://github.com/CharlyCst/miralis/blob/c25c1405bdbc1f24f95568c0e61c282710718bf5/src/virt.rs#L1064-L1067

This portion can be removed now that we have proper interrupt support, and in the test firmware `secret2` should be the same value as we put in the register, that is `0x8000000000000009`. We used to discard the interrupt bit but we should not anymore now that we support them properly. (see here: https://github.com/CharlyCst/miralis/blob/c25c1405bdbc1f24f95568c0e61c282710718bf5/firmware/mcause/main.rs#L49).

I still don't get why incremental compilation is changing the behavior here though. Is that the only test that fails?

Wesdcv:
Okay, i think i fixed most of the stuff you've mentioned, and i have a couple of questions:

- Maybe we want to have separate profiles for miralis and firmware (e.g be capable of simultaniously building miralis in release mode, but firmware in debug of vice versa)?
- Should `just gdb` have a config file as an argument as well?
- Isn't it a bit confusing to have debug option `--debug` in `just debug` and a `debug.gebug` option in config, considering they are doing different things?

Wesdcv:
> I suspect the error comes from here:
> 
> https://github.com/CharlyCst/miralis/blob/c25c1405bdbc1f24f95568c0e61c282710718bf5/src/virt.rs#L1064-L1067
> 
> This portion can be removed now that we have proper interrupt support, and in the test firmware `secret2` should be the same value as we put in the register, that is `0x8000000000000009`.

Yes, you're right, removing that return makes CI tests pass.

> I still don't get why incremental compilation is changing the behavior here though. Is that the only test that fails?

Yes, it's the only test. Moreover, only the last assertion in mcause fails - if commented out, `just test` completes successfully.

Wesdcv:
> > I suspect the error comes from here:
> > https://github.com/CharlyCst/miralis/blob/c25c1405bdbc1f24f95568c0e61c282710718bf5/src/virt.rs#L1064-L1067
> > 
> > This portion can be removed now that we have proper interrupt support, and in the test firmware `secret2` should be the same value as we put in the register, that is `0x8000000000000009`.
> 
> Yes, you're right, removing that return makes CI tests pass.

With that return removed (commented out), same assertion fails with dev profile - regardless of `incremental` value:
```
    Finished dev [optimized + debuginfo] target(s) in 0.01s
[Info  | miralis] Hart ID: 0
[Info  | miralis] misa:    0x80000000001411ad
[Info  | miralis] DTS address: 0x87e00000
[Error | miralis::virt] > Firmware: PanicInfo {
    payload: Any { .. },
    message: Some(
        assertion `left == right` failed
          left: 9223372036854775817
         right: 9,
    ),
    location: Location {
        file: "firmware/mcause/main.rs",
        line: 55,
        col: 5,
    },
    can_unwind: true,

[Error | miralis::virt] Payload panicked!
[Error | miralis::virt]   pc:    0x80200718
[Error | miralis::virt]   exits: 8
[Info  | miralis::debug] Maximal stack usage: 4072 bytes (12.24%)
error: Recipe `test` failed on line 38 with exit code 1
```
With release profile in this configuration all the tests pass just fine, as already mentioned.

CharlyCst:
The fact that it returns two different values in release and debug mode is very strange.

You can log things from firmware too, for instance we do it here in the default firmware: https://github.com/CharlyCst/miralis/blob/c25c1405bdbc1f24f95568c0e61c282710718bf5/firmware/default/main.rs#L11

You might just need to add `log = { workspace = true }` to the dependencies of the firmware to use `log`. Then you can try to log the value you read and check against in debug and release build and see if they are the same?

Wesdcv:
↑ This commit fixes the issue with assert failure
Tests pass both in release and dev profile, no more need for `incremental=true`

CharlyCst (commented):
Just one last comment/change and I think we are good for those changes :)

justfile
```diff
@@ -58,8 +59,8 @@ unit-test:
 	cargo test --features userspace -p miralis
 
 # Run Miralis
-run firmware=default:
-	cargo run --package runner -- run -v --firmware {{firmware}}
+run config=config firmware=default:
+	cargo run --package runner -- run -v --config {{config}} --firmware {{firmware}}
```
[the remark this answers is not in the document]
Wesdcv: Done!

May i ask why in `just test` build-firmware is called with `just`, but the rest of the tests are invoked through `cargo` directly?
```
# Testing with external projects
cargo run --package runner -- run --config {{qemu_virt}} --firmware opensbi
cargo run --package runner -- run --config {{qemu_virt}} --firmware zephyr --max-exits 1000000

# Test benchmark code
cargo run --package runner -- run --config {{qemu_virt_benchmark}} --firmware ecall_benchmark --benchmark

# Test firmware build
just build-firmware default {{qemu_virt}}
```

CharlyCst:
I rebased the commits on top of main and made two changes before merging:
- I removed `lto = "fat"`, I don't know why for sure but this was causing compilation errors on MacOS. I assume this is a bug on the particular toolchain we are using, but since we build with `codegen-units = 1` there should be no impact of LTO (since there is no linking step as everything is built in one go).
- I renamed the new config sections to `[target.miralis]` and `[target.firmware]`.

Thanks for the changes :)

Wesdcv:
Thank you for the rectifications!

> * I renamed the new config sections to `[target.miralis]` and `[target.firmware]`.

Is it okay that there are two `Target` structures in the runner now?

This:
https://github.com/CharlyCst/miralis/blob/36e03958eade59a1a75341ce8a02a46edcc1ab59/runner/src/artifacts.rs#L34-L37

and this:
https://github.com/CharlyCst/miralis/blob/36e03958eade59a1a75341ce8a02a46edcc1ab59/runner/src/config.rs#L94-L99

CharlyCst:
Yes I thinked about that, it is not ideal but it is easier if the name of the struct in `config.rs` is the same as in the toml config. I don't think it will cause issues in practice because `config::Target` will never be constructed outside of `config.rs` so there won't be conflict in other modules.

PR #145, opened 2024-08-15
Add support for loading OSes

### LINKED ISSUE #52: Add support for loading OSes
Currently we have good support for loading arbitrary firmware, through and optional argument to `just run`.
For instance, `just run <my_payload>` will run the provided payload (either a test firmware or a binary).

Now that we start investigating support for OSes in addition to firmware, we will eventually need a way to load OSes arbitrarily in the same way we do with firmware, ideally through one more optional argument `just run <firmware> <OS>`.
The target end goal is to be able to run a full blown Linux kernel with a single `just run opensbi vmlinux`.

While adding support for OSes, we will need to think a bit better about our naming conventions. To follow the OpenSBI conventions we will call the virtual M-mode software the firmware and the S-mode software the payload.

What needs to be done:
- [x] Complete #51
- [x] Create a new "payload" folder and populate it with a first toy OS. Note that the existing "payload" folder should have been renamed to "firmware" in #51.
- [x] Add a new linker script for payloads (i.e. OSes).
- [x] Extend the runner with the ability to build payloads (i.e. OSes), similarly to how it builds firmware.
- [x] Extend the justfile to support one extra argument for the payload.
- [ ] Add a first end-to-end test where a firmware jumps into the payload.

Note that we might need to split all of this in multiple PRs/issues.

NoeTerrier:
Payload will be placed at 0x80400000 if the path is mentioned in the just run command. It is tested with opensbi jump firmware and linux image, both built with opensbi artificat repo.

This PR aims to close #52.


CharlyCst:
Since I first wrote the issue we made a bunch of good improvements to the build system, and in particular configuration management. So I think this should be controlled from the configuration file, rather than through the `just run` command (because we might have multiple way to start the OS in the future, like OpenSBI do with `fw_payload` and `fw_jump` for instance).
We can discuss that on Monday :)

crates/abi/src/lib.rs
```diff
@@ -141,6 +141,65 @@ macro_rules! firmware_panic {
...
+            $crate::logger::init();
+
+            f();
+        }
+
+        // Defined in the linker script
+        extern "C" {
+            pub(crate) static _stack_top: u8;
+            pub(crate) static _payload_bss_start: u8;
+            pub(crate) static _payload_bss_stop: u8;
```
CharlyCst, line 198: I am wondering if we could just use the same as the firmware.

There are some minor differences, like the fact that we can't access `mhartid`, but we can work around. I'm not sure what is the best solution.

misc/linker-script-payload.x
CharlyCst, line 1: Similarly we can probably re-use the same linker script as the firmware.

At some point Miralis used a very different linker script than the firmware, but now the differences should be pretty minimal, so we might be able to use a single one for all of our use cases.

misc/riscv-unknown-payload.json
CharlyCst, line 1: Similarly we might be able to use the same as the firmware to avoid duplication. This will save compile time as the compiler can re-use libraries when building for the same target, but might be a bit more annoying because of potential name collisions between the firmware and the payload.

CharlyCst (commented):
Looks like it's taking shape :)

The one thing I am not sure about is how much duplication we want to introduce compared to what we already have with the firmware.

misc/linker-script-payload.x
NoeTerrier: There are all the same, so yes, we can use only one.

crates/abi/src/lib.rs
NoeTerrier: I don't really know how to get over the `mhartid` read for the payload. Neither of the stack size and the `panic_handler` (if we want something different for our firmware and payloads).

CharlyCst:
The PR is marked as a draft, so I'm not sure if you want a review. Remove the draft status when you want me to have a look :)

justfile
```diff
@@ -42,6 +43,8 @@ test:
 	cargo run --package runner -- run --config {{qemu_virt}} --firmware os_ecall
 	cargo run --package runner -- run --config {{qemu_virt_release}} --firmware default
 
+	cargo run --package runner -- run --config {{qemu_virt_hello_world_payload}} --firmware opensbi-jump
+
```
CharlyCst, line 47: Let's move that one just bellow `--firmware opensbi` to keep things grouped (it's a test with an external project).

runner/src/artifacts.rs
```diff
@@ -143,6 +144,26 @@ pub fn locate_artifact(name: &str) -> Option<Artifact> {
         return artifact;
     }
 
+    // Get the path to the firmware directory
```
CharlyCst, line 147: To the _payload_ directory? Not _firmware_ I assume.

runner/src/artifacts.rs
CharlyCst, line 1: Just as a side note, we will have to be careful about not creating name conflicts between firmware and payloads.

runner/src/run.rs
```diff
@@ -54,6 +57,22 @@ pub fn run(args: &RunArgs) {
             FIRMWARE_ADDR
         ));
 
+    if let Some(payload) = &cfg.target.payload {
```
CharlyCst, line 60: Maybe add a small comment above to explain what this block of code is doing (load the payload if any).

CharlyCst (approved):
Only a few nits but looks good to me :)

Congrats on getting this big refactoring working, I'll merge the PR when I get some time.

CharlyCst (approved):
Thanks! I'll merge the PR :)

PR #147, opened 2024-08-20
Improved Stack Size Configuration for Firmware

Wesdcv:
Previously, the stack size was defined as a lump value in a linker script, but now a more flexible approach was adopted.
The value is defined per-hart in a config file, similar to how it was done in miralis.

The asm part of stack size configuration is <s>copy pasted</s> heavily inspired by miralis' setup as well.

And since after those changes there were two `stack_size` options in the config file, it was decided to move the  `stack_size` option from `[platform]` section to the accrding `[target]` section, so firmware's per-hart stack size is defined in `[target.firmware]`, and miralis' in `[target.miralis]`.

### DIFF
--- a/config/example.config.toml
+++ b/config/example.config.toml
@@ -48,10 +48,6 @@ name = "qemu_virt"
 # Default to 1.
 nb_harts = 1
 
-# Size of the stack for each hart (i.e. core)
-# Default to 0x8000
-stack_size = 0x8000
-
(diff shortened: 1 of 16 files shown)


CharlyCst (approved):
Awesome, thanks for the changes :)

I spotted a bug, but it's my fault (it was probably copied from the Miralis implementation), so I'll fix it.

PR #150, opened 2024-08-21
Add a way to test a config directory

### LINKED ISSUE #148: Check all config files in the CI
So far we manually check a few config files in the CI, see: https://github.com/CharlyCst/miralis/blob/eb8d91c3c9a06437b954b01ea1f6517ee51a4c3c/justfile#L28-L32

The problem is we lost track of all the configs, and as a result not all of the config files are tested and guaranteed to work at a given commit.

To ensure all of our config files are always working we should automatically check all configs in the `./config` directory. For that we could add a new flag to the Runner's `check-config` command, let's call it `--walkdir` for lack of a better name, that walks the directory at the given path and checks all the files ending in `.toml`.
Once implemented we can modify the CI to check all configs in `./config` and ensure they are always up-to-date.

To implement the recursive directory search we can use the popular and well maintained [`walkdir` crate](https://docs.rs/walkdir/latest/walkdir/).

NoeTerrier:
The argument config for the check-config command of the runner can now be the path of a folder. In the latter case, the runner will walk the directory, filter toml file and test them.

Close #148

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -310,6 +310,7 @@ dependencies = [
  "config_helpers",
  "serde",
  "toml",
+ "walkdir",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


CharlyCst (approved):
Thanks for the changes! That's definitely going to catch invalid configs in the future :)

PR #153, opened 2024-08-23
Initialization of the BSS is done by the boot hart.

### LINKED ISSUE #151: Multicore-support for BSS initialization
In the current implementation all threads are zeroing out the BSS, which can lead to race condition and loss of data.

https://github.com/CharlyCst/miralis/blob/332ed085096d624f59beb2fb3afa105b579b043b/src/arch/metal.rs#L771-L781

To allow for correct execution on multicore platforms we need to ensure that:
- A single hart zeroes-out the BSS section
- No hart exits the assembly boot code before the BSS section is zeroed out.

In practice that means that we need to pick a boot hart and have it zero-out all of the BSS while the other harts are waiting.
We already have a concept of boot hart, so we can re-use it to choose which core should boot first rather than doing a race.
For all other harts we need to setup a synchronization barrier. The easiest option is to wait on an atomic, the fancy version would be to do a `wfi` but that makes the implementation drastically more complication so I would suggest to stick to the basic for now.

NoeTerrier:
All other hart wait for the primary hart to initialize the BSS before jumping to main. This avoids race condition on the BSS section, which should be zero.

This is maybe missing a test. How could it be tested ?

Close #151

### DIFF
--- a/src/arch/metal.rs
+++ b/src/arch/metal.rs
@@ -5,7 +5,7 @@ use core::{ptr, usize};
 
 use super::{Architecture, Csr, MCause, Mode, RegistersCapability, TrapInfo};
 use crate::arch::{mstatus, HardwareCapability, PmpGroup};
-use crate::config::TARGET_STACK_SIZE;
+use crate::config::{PLATFORM_BOOT_HART_ID, TARGET_STACK_SIZE};
 use crate::virt::VirtContext;
 use crate::{_bss_start, _bss_stop, _stack_start, main};
 
@@ -734,6 +734,7 @@ unsafe fn write_pmpcfg(index: usize, pmpcfg: usize) {
 
 global_asm!(
 r#"
+.attribute arch, "rv64imac"
 .align 4
 .text
 .global _start
@@ -769,8 +770,13 @@ stack_fill_loop:
 stack_fill_done:
 
     // Now we need to zero-out the BSS section
-    // TODO: this is currently done by all cores, but we should have some sort of synchronization
-    // to avoid race conditions.
+    // Only the boot hart set the BSS section to avoid race condition.
+
+    csrr t0, mhartid         // Our current hart ID
+    ld t2, {boot_hart_id}    // Boot hart ID
+    ld t3, __boot_bss_set    // Shared boolean, set to 1 to say to other harts that the BSS is not initialized yet
+    bne t0, t2, wait_bss_end // Only the boot hart initializes the bss
+
     ld t4, __bss_start
     ld t5, __bss_stop
 zero_bss_loop:
@@ -779,6 +785,19 @@ zero_bss_loop:
     addi t4, t4, 8
     j zero_bss_loop
 zero_bss_done:
+
+    // Say to other harts that the initialization is done.
+    // This is atomic and memory is ordered in a way that the 
+    // initialization will then be visible as soon as the
+    // boolean is reset. .aqrl means that it is done sequentially.
+    amoswap.w.aqrl x0, x0, (t3)
+    j end_wait
+    
+    // Wait until initialization is done
+wait_bss_end:
+    lw t2, (t3)
+    bnez t2, wait_bss_end
+end_wait:
 
     // And finally we load the stack pointer into sp and jump into main
     mv sp, t1
@@ -793,12 +812,16 @@ __bss_start:
     .dword {bss_start}
 __bss_stop:
     .dword {bss_stop}
+__boot_bss_set:
+    .dword {boot_bss_set}
 "#,
     main = sym main,
     stack_start = sym _stack_start,
     stack_size = sym STACK_SIZE,
     bss_start = sym _bss_start,
     bss_stop = sym _bss_stop,
+    boot_hart_id = sym BOOT_HART_ID,
+    boot_bss_set = sym BOOT_BSS_SET,
 );
 
 // NOTE: We need to use a static here because constant in `asm!` blocks are not yet supported.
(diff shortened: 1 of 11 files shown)


src/arch/metal.rs
```diff
@@ -809,6 +834,12 @@ __bss_stop:
 // https://github.com/rust-lang/rust/issues/93332
 static STACK_SIZE: usize = TARGET_STACK_SIZE;
 
+// Boot hart ID
+static BOOT_HART_ID: usize = PLATFORM_BOOT_HART_ID;
+
+// Boolean to synchronized harts
+static BOOT_BSS_SET: bool = true;
```
CharlyCst, line 841: Rust does not give layout guarantees for booleans. So this code might break when upgrading the compiler if the layout changes for some reason. Also big/little endian change the conversions between boolean and integers.

So you should pick a well defined integer size and initialize it with a proper layout. Then be sure that your load matches the size of the integer type you picked.

src/arch/metal.rs
```diff
@@ -769,8 +770,15 @@ stack_fill_loop:
...
-    // TODO: this is currently done by all cores, but we should have some sort of synchronization
-    // to avoid race conditions.
+    // Only the boot hart set the BSS section to avoid race condition.
+
+    csrr t0, mhartid       // Our current hart ID
+    ld t2, {boot_hart_id}  // Boot hart ID
+    bne t0, t2, wait_bss_end // Only the boot hart initializes the bss
+    
+    // Shared boolean, set to 1 to say to other harts that the BSS is not initialized yet
+    ld t3, __boot_bss_set
```
CharlyCst, line 780: You load the address of the boolean _after_ the jump of the non-boot harts. So the non-boot harts are truing to read from `t3` but `t3` might contain some invalid address.

You should load it before to ensure all threads have the address of the variable in `t3`.

src/arch/metal.rs
```diff
@@ -780,6 +788,19 @@ zero_bss_loop:
...
+    // Say to other harts that the initialization is done.
+    // This is atomic and memory is ordered in a way that the 
+    // initialization will then be visible as soon as the
+    // boolean is reset.
+    amoswap.w.rl x0, x0, (t3)   
+    j end_wait
+    
+    // Wait until initialization is done
+wait_bss_end:
+    lw t1, (t3)
```
CharlyCst, line 800: Here you need an atomic too, probably with acquire semantic (you should always pair a release with an acquire to guarantee ordering, at least on other architecture I know about).

CharlyCst: Actually I realize RISC-V might not call those atomics, but maybe just loads with acquire semantic?

CharlyCst (changes requested):
Thanks for the changes!

I think there are some issued in the boot assembly, maybe that's what causing issues later on?

src/arch/metal.rs
NoeTerrier: Has `usize` a well-defined integer size?

CharlyCst: It is platform dependent (32 or 64 bits), but yes the layout is specified ^^

src/arch/metal.rs
NoeTerrier: I can replace `amoswap.w.rl` by `amoswap.w.aqrl`. This means that the operation will be done in a sequential order in respect to the BSS zeroing. It will be visible by other hart in a sequential order too.

We do not need the other harts to acquire a lock, but only to pull the boolean until it is set to 0 by the boot hart.
The boolean is set to 1 by default, and cleared only by hart 0. We need to ensure that the BSS zeroing is done and visible before the boolean clear. So sequential ordering of the boolean clear should work, right?

Spec:

> To help implement multiprocessor synchronization, the AMOs optionally provide release consis-
tency semantics. If the aq bit is set, then no later memory operations in this RISC-V hart can be
observed to take place before the AMO. Conversely, if the rl bit is set, then other RISC-V harts will
not observe the AMO before memory accesses preceding the AMO in this RISC-V hart. Setting
both the aq and the rl bit on an AMO makes the sequence sequentially consistent, meaning that it
cannot be reordered with earlier or later memory operations from the same hart.

CharlyCst: The acquire and release semantic are local to a hart, by themselves they do not change what other harts can see.

Here the release semantic means that every operations before the AMO will be ordered before the AMO, while the acquire means that no operation that happens after the AMO will be ordered before the AMO. But you have to think of that as the other in which the hart will submit the changes to the memory system (caches, etc...).

To see a consistent ordering between two harts we need two things:
1. The hart doing the writes needs to decide on its own ordering (the order in which it submits changes to the memory subsystem).
2. The hart doing the reads needs to check if there were updated on the value it reads, so it needs to ask the memory subsystem if anything changed and in which order.

You are doing (1), but not (2), because you don't use a fence or atomic operation to read the value from memory.
Feel free to ask if anything is unclear :)

CharlyCst (commented):
See the discussion on atomics, feel free to ask if you want to discuss that particular point :)

src/main.rs
```diff
@@ -55,14 +55,7 @@ mod userspace_linker_definitions {
...
 use userspace_linker_definitions::*;
 
-pub(crate) extern "C" fn main(hart_id: usize, device_tree_blob_addr: usize) -> ! {
-    // For now we simply park all the harts other than the boot one
-    if hart_id != Plat::get_primary_hart() {
-        loop {
-            Arch::wfi();
-            core::hint::spin_loop();
-        }
-    }
```
CharlyCst, line 65: We might now want to delete this in this PR, as the rest of the code is not ready for multi-core I assume.

I'll handle it before merging :)

CharlyCst (approved):
Thanks, looks good. I'll merge the PR when I get some time, thanks for the changes!

PR #158, opened 2024-08-27
Implement logger enabled function.

NoeTerrier:
This will avoid computing argument of logging if log level is not enabled.

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -35,39 +35,36 @@ impl Logger {
         return false;
     }
 
-    fn filter_by_module(&self, record: &Record) -> bool {
+    fn filter_by_module(&self, metadata: &Metadata) -> bool {
         let mut specific_module_enabled: bool = false;
 
-        specific_module_enabled |= record.metadata().level() <= LevelFilter::Trace
-            && Self::contains_target(config::LOG_TRACE, record.target());
-        specific_module_enabled |= record.metadata().level() <= LevelFilter::Debug
-            && Self::contains_target(config::LOG_DEBUG, record.target());
-        specific_module_enabled |= record.metadata().level() <= LevelFilter::Info
-            && Self::contains_target(config::LOG_INFO, record.target());
-        specific_module_enabled |= record.metadata().level() <= LevelFilter::Warn
-            && Self::contains_target(config::LOG_WARN, record.target());
-        specific_module_enabled |= record.metadata().level() <= LevelFilter::Error
-            && Self::contains_target(config::LOG_ERROR, record.target());
+        specific_module_enabled |= metadata.level() <= LevelFilter::Trace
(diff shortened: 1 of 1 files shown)


src/logger.rs
```diff
@@ -58,9 +58,17 @@ impl Logger {
...
+        // We filter depending on the current log level.
+        metadata.level()
+            <= match config::LOG_LEVEL {
+                Some("trace") => LevelFilter::Trace,
+                Some("debug") => LevelFilter::Debug,
+                Some("info") => LevelFilter::Info,
+                Some("warn") => LevelFilter::Warn,
+                Some("error") => LevelFilter::Error,
+                Some(&_) | None => LevelFilter::Off,
+            }
```
CharlyCst, line 71: Unfortunately it's not that simple, because we can enable per-module level. You have to use the `filter_by_global_level` (which does what you have here) and `filter_by_module` functions and do an or between the two (see `log` implementation bellow).

CharlyCst (changes requested):
See comment, we also need to check for per-module logging level. We already have functions for that though :)

CharlyCst:
To avoid creating a new commit you can use `git commit --amend` instead of pushing a new one :)

src/logger.rs
NoeTerrier: I see, a problem here is that the `enabled` take a `&Metadata` object and not a `&Record`. And so I have no idea how to get the target.

CharlyCst: Hmmm, actually it looks like those two `filter_by_*` functions only use the record from the metadata, so we could change their signatures to take a record (and pass it from `metadata.record` in `log`).

NoeTerrier: Actually, I think it is the opposite: metadata come from the record, and trait `log::Log` define `enabled` function using only metadata.

CharlyCst (commented):
I think my previous answer was a bit confusing, fell free to ask for clarification :)

src/logger.rs
```diff
@@ -58,9 +45,22 @@ impl Logger {
...
+        specific_module_enabled |= metadata.level() <= LevelFilter::Debug
+            && Self::contains_target(config::LOG_DEBUG, metadata.target());
+        specific_module_enabled |= metadata.level() <= LevelFilter::Info
+            && Self::contains_target(config::LOG_INFO, metadata.target());
+        specific_module_enabled |= metadata.level() <= LevelFilter::Warn
+            && Self::contains_target(config::LOG_WARN, metadata.target());
+        specific_module_enabled |= metadata.level() <= LevelFilter::Error
+            && Self::contains_target(config::LOG_ERROR, metadata.target());
+
+        specific_module_enabled
```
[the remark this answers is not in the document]
NoeTerrier: Oh I see, I hope it's good now haha

CharlyCst (approved):

PR #159, opened 2024-08-27
Avoid MEIE and MSIP to be set in the virtual context

### LINKED ISSUE #155: Linux multi hart wake up fails
**Context:** 
When booting Linux with opensbi on multi harts, Linux fails to wake up cpus. 

```
[    1.085495] CPU1: failed to come online
[    2.131926] CPU2: failed to come online
[    3.186053] CPU3: failed to come online
```

**What happen:**

- Linux calls `smpboot.c/__cpu_up()`, 
  - Which calls `start_secondary_cpu` which finally produce an ecall
   `sbi_ecall(SBI_EXT_HSM, SBI_EXT_HSM_HART_START, hartid, saddr, priv, 0, 0, 0);`  
- The ecal is well received by opensbi, which set up an interrupt (MSIP) via the CLINT for the hart ´hartid´. 
Then, it mret to Linux.
- Linux `wait_for_completion_timeout` for 1s, when the timeout hits, it checks if the cpu is online with a mask.
But the cpu is not online.

**Expected**
The cpus should be all online before the timeout.

**Reproduce**
1. checkout commit `da3533f07114d9e03496fc41e1cc6f656693cab5` on branch `151-multicore-support-for-bss-initialization`
2. set number of hart more than one in the config
3. `just run linux`

NoeTerrier:
To wake up harts, firmware might use a machine software interrupt (`MSI`) or am machine external interrupt (`MEI`). These interrupts need to be fetched from hardware (`mip`) for each virtual read of `vmip`, as they can occur asynchronously with the execution of the firmware. During a world switch, we need to take care that these interrupts are not installed in the virtual `vmip`, to avoid having an interrupt that can't be cleared by firmware in the virtual context.

Close  #155

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -238,3 +238,7 @@ Another thing to consider is the following scenario:
 
 If the firmware wants to read the `mip` register after cleaning `vmip.SEIP`, and we don't sync `vmip.SEIP` with `mip.SEIP`, it can't know if there is an interrupt signal from the interrupt controller as the CSR read will be a logical-OR of the signal and `mip.SEIP` (which is one), and so always 1. If vmip.SEIP is 0, CSR read of mip.SEIP should return the interrupt signal.
 Then, we need to synchronize vmip.SEIP with mip.SEIP.
+
+### Mutliharts interrupt
+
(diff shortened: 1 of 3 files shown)


CharlyCst:
Thank for the fix! So with this we have multi-core properly running? Sounds awesome :)

We will have to start thinking about how to test this properly 🤔 the easiest is probably to run Linux...

Issue #154, opened 2024-08-25
``mtime`` returns random values while reading from U-mode


francois141:
When attempting to read the mtime register in userspace, it appears to return random values. Below is an example of the output that demonstrates this issue.

```bash
[Warn  | miralis::virt] > Reading from mtime 9673451624126742528
[Warn  | miralis::virt] > Reading from mtime 24489422485454848
[Warn  | miralis::virt] > Reading from mtime 8890951188871118848
[Warn  | miralis::virt] > Reading from mtime 2193254118041059328
[Warn  | miralis::virt] > Reading from mtime 6088586320739827712
[Warn  | miralis::virt] > Reading from mtime 13153608236177293312
```

To reproduce the issue, execute the following commands:

```bash
git checkout mtime-issue; cargo run --package runner -- run --config {{qemu_virt}} --firmware interrupt
```

CharlyCst:
Interesting, it seems to be working on my machine:

```
[Warn  | miralis::virt] > mtime: 43660
[Warn  | miralis::virt] > mtime: 47030
[Warn  | miralis::virt] > mtime: 48020
[Warn  | miralis::virt] > mtime: 48990
[Warn  | miralis::virt] > mtime: 50020
[Warn  | miralis::virt] > mtime: 50990
```

Which version of QEMU are you using (`qemu-system-riscv64 --version`)? I am on  9.0.1.

francois141:
I use an older version

```bash
fran-ois-costa@fran-ois-costa-1-01:~$ qemu-system-riscv64 --version
QEMU emulator version 8.0.4 (Debian 1:8.0.4+dfsg-1ubuntu3.23.10.5)
Copyright (c) 2003-2022 Fabrice Bellard and the QEMU Project developers

```

francois141:
@CharlyCst It seems that it doesn't read the CLINT at all. I will investigate that

CharlyCst:
Wow, ok this is getting stranger ^^ do you two computers have the same qemu version?

francois141:
The other computer has : ``QEMU emulator version 9.0.2``

CharlyCst:
Could this be related to the bug in #161?

I am not super enthusiastic about requiring a minimal QEMU version, it will make onboarding harder and unless we can get reasonably confident that the bug is in QEMU (could be, who knows) I would rather suspect that we are doing something wrong somewhere... Better fix the bug than hide it :)

CharlyCst:
[...]

PR #161, opened 2024-08-28
Test virtual device & test firmware

francois141:


src/host.rs
```diff
@@ -16,7 +16,7 @@ pub struct MiralisContext {
     /// Hardware capabilities of the core (hart).
     pub hw: HardwareCapability,
     /// List of devices with PMP
-    pub devices: [device::VirtDevice; 1],
+    pub devices: [device::VirtDevice; 2],
```
francois141, line 19: @CharlyCst Do we want to have a test device on each platform?

CharlyCst: Let's say yes for now, I don't expect it to add much overhead in terms of space or performance.
In the future we will need to have a better way of describing platforms and their devices, but we don't have enough devices to worry about it yet :)

src/device/clint.rs
```diff
@@ -0,0 +1,98 @@
+use spin::Mutex;
+
+use crate::device::{DeviceAccess, Width};
+use crate::driver::clint::{
+    MSIP_OFFSET, MSIP_WIDTH, MTIMECMP_OFFSET, MTIMECMP_WIDTH, MTIME_OFFSET,
+};
+use crate::driver::ClintDriver;
+// ————————————————————————————— Virtual CLINT —————————————————————————————— //
```
CharlyCst, line 8: Style nit: we usually add a space before comment separators ^^
```suggestion

// ————————————————————————————— Virtual CLINT —————————————————————————————— //
```

src/device/mod.rs
```diff
@@ -0,0 +1,64 @@
+//! Base device classes
+
+pub mod clint;
+pub mod tester;
+// ———————————————————————————— Virtual Devices ————————————————————————————— //
```
CharlyCst, line 5: Same style remark, we add a space before here
```suggestion

// ———————————————————————————— Virtual Devices ————————————————————————————— //
```

src/device/tester.rs
```diff
@@ -0,0 +1,67 @@
+use core::sync::atomic::{AtomicUsize, Ordering};
+
+use crate::device::{DeviceAccess, Width};
+
+// ————————————————————————————— Virtual Test Device —————————————————————————————— //
+
+pub const TEST_DEVICE_SIZE: usize = 0x8;
+
+#[derive(Debug)]
+pub struct VirtTestDevice {
```
CharlyCst, line 10: Maybe we can have a small doc comment here to explain what the purpose of this device is (testing).

CharlyCst (commented):
The code looks good to me :)

Does the test pass? If yes we can add it to `just test` to run it in the CI

francois141:
@CharlyCst Do you know what could go wrong here....?

Capture d’écran 2024-08-29 à 15 53 18

CharlyCst:
Wow interesting, both numbers are printed as being exactly the same?

Assuming the device emulation indeed returns the right value, it is possible that somehow we are invoking undefined behavior here, maybe by polluting some registers... I would try to see if the state is corrupted when coming back from device emulation.
One way to do that would be to look at the assemble and step through it with GDB (one breakpoint before the read, one breakpoint after and see if all registers make sense). Another option is to try to rewrite the assert as an `if` branch and a panic and see if it still crashed, if that is the case looking at the assembly of the `if` should be simpler than the assert and GDB can probably give an answer.

CharlyCst:
Hmmm, something strange is going on here, I am not sure but it might be related to #154.
I'd like to investigate a bit before merging this, I'll take a look and see if I find anything suspicious.

CharlyCst:
The issue seems to be a bug in our load emulation (`lw` in particular).
The spec says that `lw` should perform a sign extension, which is not working properly in our emulation. As a result the test tries to compare `0xffffffffdeadbeef` (loaded from the `0xdeadbeef` hardcoded constant) and `0xdeadbeef` from reading the test device. The write should sign extend to `0xffffffffdeadbeef` so that the comparison can be done properly.

I'll investigate the root cause a bit further.

CharlyCst:
So turns out the test driver caught a bug in our load emulation ^^
I pushed a fix, the test passes properly with `0xdeadbeef` now.

PR #164, opened 2024-08-29
Remove parking of other harts

NoeTerrier:
Before, harts except boot hart were put in an idle state. We can remove it now.
We can check if the bug of VisionFive2 is still here. (#122)


CharlyCst:
Thanks, I merged the changes :)

PR #165, opened 2024-08-29
Add qemu section to the config

### LINKED ISSUE #157: Add [qemu] section to config
So far we don't really have a way to configure QEMU for running our tests. For instance there is no way to choose the cpu model or board to emulate. This is problematic because different person can end up with different results (and in particular the CI can use different cpu features over time).

The goal of this issue is to add a new `[qemu]` config with two fields:
- `machine`, which should be a QEMU machine (e.g. `virt`, `sifive_u`, or others).
- `cpu`,  which specify the CPU model (e.g. `sifive-u54`, or default to not specifying anything)

Once we have this new section we can add Linux to the CI using the `sifive-u54` CPU core, yeay 🥳

NoeTerrier:
Qemu parameter "machine" and "cpu" could be set to choose machine and cpu of qemu. If cpu is set to none, no cpu is provided and a default one is used.

Close #157

### DIFF
--- a/config/example.config.toml
+++ b/config/example.config.toml
@@ -52,6 +52,18 @@ nb_harts = 1
 # Default to 0
 boot_hart_id = 0
 
+[qemu]
+
+# Qemu machine (virt, sifive_u, spike...) 
+# Full list given by qemu-system-riscv64 --machine help
+# Default to "virt"
+machine = "virt"
+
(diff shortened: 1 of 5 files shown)


CharlyCst (approved):
Wow, huge milestone to run Linux in the CI 🥳

Thanks for the changes!

Issue #67, opened 2024-05-28
Try running RustSBI?


CharlyCst:
We already support running OpenSBI on the CI and locally (as a downloadable artifact). Can we also run RustSBI?

There seems to be a QEMU port [here](https://github.com/rustsbi/rustsbi-qemu), this seems like a good point to start investigation. If successful, we could add RustSBI to the CI like we did with OpenSBI.

Interestingly, they have a "bench-kernel" folder, maybe that could be useful to us as a simple benchmark to compare with their baseline?

CharlyCst:
I was wondering, do you think we could easily add RustSBI as an artifact once we complete #52?
My understanding is that the main blocker was the need for loading a kernel in memory at some address, which we should be able to do afterward.

NoeTerrier:
Improvement since:

```
[Info  | miralis] Hello, world!
[Info  | miralis] Platform name: QEMU virt
[Info  | miralis] Hart ID: 0
[Info  | miralis] misa:    0x80000000001411ad
[Info  | miralis] vmisa:   0x8000000000141101
[Info  | miralis] mstatus: 0xa00000000
[Info  | miralis] DTS address: 0x87e00000
[Info  | miralis] Preparing jump into firmware
[rustsbi] RustSBI version 0.4.0-alpha.1, adapting to RISC-V SBI v2.0.0
.______       __    __      _______.___________.  _______..______   __
|   _  \     |  |  |  |    /       |           | /       ||   _  \ |  |
|  |_)  |    |  |  |  |   |   (----`---|  |----`|   (----`|  |_)  ||  |
|      /     |  |  |  |    \   \       |  |      \   \    |   _  < |  |
|  |\  \----.|  `--'  |.----)   |      |  |  .----)   |   |  |_)  ||  |
| _| `._____| \______/ |_______/       |__|  |_______/    |______/ |__|
[rustsbi] Implementation     : RustSBI-QEMU Version 0.2.0-alpha.3
[rustsbi] Platform Name      : riscv-virtio,qemu
[rustsbi] Platform SMP       : 2
[rustsbi] Platform Memory    : 0x80000000..0x88000000
[rustsbi] Boot HART          : 0
[rustsbi] Device Tree Region : 0x87e00000..0x87e012c6
[rustsbi] Firmware Address   : 0x80200000
[rustsbi] Supervisor Address : 0x80400000
[rustsbi] pmp01: 0x00000000..0x80000000 (-wr)
[...]

PR #168, opened 2024-08-29
Add support for vectored mtvec

### LINKED ISSUE #167: Support `mtvec` vectored mode
Currently, only direct mode is implemented for redirecting traps to a trap handler.

The second mode is vectored mode. The mode is stored inside the two least significant bits inside `mtvec`.

Spec:
> 0: direct : All traps set pc to BASE.
> 1: vectored : Asynchronous interrupts set pc to BASE+4×cause.

RustSBI use vectored mode. #67

NoeTerrier:
Modify a bit `virt.rs` and `mod.rs` to support vectored `mtvec`. 
This is tested with RustSBI and a new test in the firmware folder.

We can probably create a repo for RustSBI and add it to the CI if we want now. #67 

Close #167

### DIFF
--- a/firmware/vectored_mtvec/main.rs
+++ b/firmware/vectored_mtvec/main.rs
@@ -1,0 +1,103 @@
+#![no_std]
+#![no_main]
+
+use core::arch::{asm, global_asm};
+use core::hint::spin_loop;
+
+use miralis_abi::{failure, setup_binary, success};
+
+setup_binary!(main);
+
+fn main() -> ! {
+    // Set mtvec to vectored mode at base _raw_interrupt_trap_handler
+    // Then produce an interrupt (MTI) to fall inside the vector
+    // The redirection should end in success_trap_handler
+    unsafe {
+        let _raw_interrupt_trap_handler = _raw_interrupt_trap_handler as usize | 0b1;
+
+        asm!(
+            "csrw mtvec, {handler}",       // Setup trap handler
+            "csrw mideleg, 0",
+            "csrs mstatus, {mstatus_mie}", // Enable interrupts (MIE)
+            "csrs mie, {mtie}",            // Enable machine timer interrupt (MTIE)
+            handler = in(reg) _raw_interrupt_trap_handler,
+            mstatus_mie = in(reg) 0x8,
+            mtie = in(reg) 0x80,
+        );
+    }
+
+    // Setup a timer deadline in the past to trap directly
+    set_mtimecmp_future_value();
+
+    for _ in 0..10_000 {
+        spin_loop()
+    }
+
+    // The trap handler should exit, if we reach that point the handler did not do its job
+    log::error!("Firmware didn't trapped!");
+    failure()
+}
+
+// ———————————————————————————— Timer Interrupt ————————————————————————————— //
+
+const CLINT_BASE: usize = 0x2000000;
+const MTIMECMP_OFFSET: usize = 0x4000;
+
+// Set mtimecmp value in the future
+fn set_mtimecmp_future_value() {
+    let future_time = 0;
+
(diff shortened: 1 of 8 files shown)


NoeTerrier:
Really strange, the tests all pass on my laptop but not in the CI

```
 cargo run --package runner -- run --config ./config/test/qemu-virt.toml --firmware vectored_mtvec
    Finished dev [optimized + debuginfo] target(s) in 0.02s
     Running `target/debug/runner run --config ./config/test/qemu-virt.toml --firmware vectored_mtvec`
Running Miralis with 'vectored_mtvec' firmware
"/home/runner/work/miralis/miralis/target/riscv-unknown-miralis/debug"
    Finished dev [optimized + debuginfo] target(s) in 0.03s
"/home/runner/work/miralis/miralis/target/riscv-unknown-unprivileged/debug"
   Compiling vectored_mtvec v0.1.0 (/home/runner/work/miralis/miralis/firmware/vectored_mtvec)
    Finished dev [optimized + debuginfo] target(s) in 0.10s
[Info  | miralis] Hart ID: 0
[Info  | miralis] misa:    0x800000000014112d
[Info  | miralis] DTS address: 0x87000000
[Error | miralis::virt] Firmware or payload panicked!
[Error | miralis::virt]   pc:    0x80200b3c
[Error | miralis::virt]   exits: 4
[Info  | miralis::debug] Maximal stack usage: 4072 bytes (12.24%)
error: Recipe `test` failed on line 44 with exit code 1
Error: Process completed with exit code 1.
```

firmware/vectored_mtvec/main.rs
```diff
@@ -0,0 +1,98 @@
...
+            mtie = in(reg) 0x80,
+        );
+    }
+
+    // Setup a timer deadline in the past to trap directly
+    set_mtimecmp_future_value();
+
+    // The trap handler should exit, if we reach that point the handler did not do its job
+    log::error!("Firmware didn't trapped!");
+    failure()
```
CharlyCst, line 33: Hmmm technically there could be a delay because it is MMIO (Memory Mapped I/O) which is asynchronous, maybe we should spin a bit with a loop. For instance `10_000` iterations with a [`core::hint::spin_loop`](https://doc.rust-lang.org/stable/core/hint/fn.spin_loop.html#).

src/arch/mod.rs
```diff
@@ -345,3 +345,31 @@ pub mod mie {
...
+
+// ———————————————————— Machine Trap-Vector Base-Address ———————————————————— //
+
+#[allow(unused)]
+pub mod mtvec {
+    /// Constant to filter out MODE bits of mtvec
+    pub const MODE_FILTER: usize = 0b11;
+
+    /// Constant to filter out BASE bits of mtvec
+    pub const BASE_FILTER: usize = usize::MAX - MODE_FILTER;
```
CharlyCst, line 357: Si indeed that does work, but it is kind of unusual to do bit manipulations with subtractions. In particular it works because all of the bits of `usize::MAX` are one, and if someone change it in the future this will break so it is kind of fragile.

What I would use here is the usual `usize::MAX & !MODE_FILTER` which does the same thing but is a bit more robust and commonly used.

CharlyCst (approved):
Just some small nits but looks good to me :)

Feel free to fix them but I can also do it a bit later when I get some time and then merge.

NoeTerrier:
> It seems to work on the CI now, do you know what the issue was?

Yep, the interrupt should be enabled by a timer (I think it is the only way m-mode can set up an interrupt for itself). On my machine, writing to mip at this bit was allowed so it trapped, but I think it is read-only on the CI, so the bit was not set and so no trap.

firmware/vectored_mtvec/main.rs
NoeTerrier: I see, yes good idea

PR #169, opened 2024-08-29
Add RustSBI to the CI.

### LINKED ISSUE #67: Try running RustSBI?

NoeTerrier:
Two binaries, that work together, are added. The first one is the [rustsbi](https://github.com/rustsbi/rustsbi-qemu) binary, and the second one is a test-kernel that can be used as a payload with it.

We need to wait for #168 to be merged.

Close #67

### DIFF
--- a/config/test/qemu-rustsbi-test-kernel.toml
+++ b/config/test/qemu-rustsbi-test-kernel.toml
@@ -1,0 +1,56 @@
+# A test configuration to run on QEMU virt platform
+
+[log]
+level = "info"
+color = true
+
+[debug]
+max_firmware_exits = 2000
+
+[vcpu]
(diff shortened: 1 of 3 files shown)


CharlyCst (approved):
Cool, more tests :)

Thanks for the changes!

PR #170, opened 2024-08-29
Refactored load&store instructions

Wesdcv:
**Summary**

Previously, our codebase was suboptimal for the organization of load and store instructions. Each ISA instruction had its own separate enum field, resulting in redundancy and maintenance challenges. However, many load and store instructions share their meaningful fields.

**Key Changes**

- Common Structure: By identifying a common structure for load and store instructions, it was possible to reduce the number of enum variants—from 12 down to just 2. This streamlined approach ensures scalability even if we decide to add support for floating-point or double-precision load and store instructions in the future.
- Refactored Handlers: Sequentially refactored the handlers for load and store operations, consolidating them into just 2 functions. This reduction in repetition simplifies the codebase and improves maintainability.
- Unsigned Load Support: Adding support for unsigned loads became trivial after these changes, so i decided to leave it on the same branch.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -1,5 +1,5 @@
 //! RISC-V instruction decoder
-use crate::arch::{Csr, Register};
+use crate::arch::{Csr, Register, Width};
 use crate::platform::{Plat, Platform};
 
 const OPCODE_MASK: usize = 0b1111111 << 0;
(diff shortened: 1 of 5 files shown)


src/decoder.rs
```diff
@@ -49,77 +49,22 @@ pub enum Instr {
...
-    /// Load Double-Word (register-based)
-    Ld {
+    /// Load (register-based)
+    Load {
         rd: Register,
         rs1: Register,
         imm: isize,
+        len: usize,
+        compressed: bool,
+        unsigned: bool,
```
CharlyCst, line 59: A good convention to name boolean variables (across multiple languages, not just in Rust) is to start with `is_`. For instance `is_compressed` and `is_unsigned`. It works well with `if` statements (e.g. `if load.is_compressed { ... }` and it makes boolean easy to identify when reading the code.

src/decoder.rs
```diff
@@ -179,22 +124,49 @@ fn decode_c_reg_based(raw: usize) -> Instr {
         0b111 => {
             let rs2 = rd_rs2;
             let imm = 0b000 | (raw >> 7) & 0b111000 | ((raw << 1) & 0b11000000);
-            Instr::CSd { rs1, rs2, imm }
+
```
CharlyCst, line 127: Small style nit: you don't have a while line here in the other cases, let's keep things consistent :)

```suggestion
```

src/decoder.rs
```diff
@@ -179,22 +124,49 @@ fn decode_c_reg_based(raw: usize) -> Instr {
...
+            Instr::Store {
+                rs2,
+                rs1,
+                imm: (imm * 8).try_into().unwrap(),
+                len: (64),
+                compressed: (true),
```
CharlyCst, line 133: Not really important, but any reasons for the parenthesis here?

src/decoder.rs
```diff
@@ -223,10 +195,72 @@ fn decode_load(raw: usize) -> Instr {
...
+        0b110 => Instr::Load {
+            rd,
+            rs1,
+            imm,
+            len: (32),
+            compressed: (false),
+            unsigned: (true),
+        },
+        // On a RV64 ISA this should never be the case,
+        // But for the sake of completeness we might decode it as well
```
CharlyCst, line 255: Which instruction is that? 128 bits load?
If it is we should return `Instr::Unknown`, otherwise we will silently emulate it and not be compliant with the spec.

src/decoder.rs
```diff
@@ -49,77 +49,22 @@ pub enum Instr {
...
     Sret,
     Vfencevma,
-    /// Load Double-Word (register-based)
-    Ld {
+    /// Load (register-based)
+    Load {
         rd: Register,
         rs1: Register,
         imm: isize,
+        len: usize,
```
CharlyCst, line 57: We already have an enum with the possible cases here.

https://github.com/CharlyCst/miralis/blob/ad9f737ac97eab3242cffc5b4f38351398c4e534/src/device.rs#L9-L19

Currently it is in Device, maybe we could move it to `decoder.rs` if that makes more sense, but it would be better to re-use it throughout the whole code base rather than using different way of representing the width.

CharlyCst (commented):
Thanks for the refactoring! This is indeed going to make things simpler :)

I just have a few comment, mostly style except for the `len` where I think we should use the same abstraction throughout the codebase.

src/decoder.rs
Wesdcv: Okay, won't it make more sense then to move the `Width` enum to `utils.rs` or other general-purpose module? I anticipate the enum to be used to some extent in `decoder`, `device`, `driver`, `metal` and `virt` modules.

CharlyCst: Then yes `util.rs` works, `arch/mod.rs` can work too, we already have some generic architecture definitions there, both sounds good to me :)

src/decoder.rs
Wesdcv: It would be an LDU instruction, it's a part of RV128I set.
Sure, i'll remove that match option.

src/decoder.rs
Wesdcv: Not really, my IDE put them in place as a suggestion and `just test` didn't deem them as unnecessary.
I can remove them, np :)

CharlyCst (dismissed):
Awesome, looks all good to me :)

Thanks for getting this done! It's going to make things easier ^^

I'll merge it when I get some time, there might be some conflicts with the PRs from François, I'll handle those

PR #171, opened 2024-08-30
Refactor config.rs to be easier to add more param

NoeTerrier:
Refactor code of `config.rs`, New struct `EnvVars` which can be used to append optional environment parameters easily.


CharlyCst (approved):
Thanks for the change :)

Indeed, this looks much better!

PR #173, opened 2024-08-30
Decode `sfence.vma` with registers

### LINKED ISSUE #172: Time command in Linux shell fails.
**Context:**
When running linux `just run linux-shell`, I eventually end up in an error, or an infinite loop. 

It appears to happen only when using mutlicores.

**What happen:**
Crash or infinite waiting.

```
Boot took  seconds

Hello from Linux!

/bin/sh: can't access tty; job control turned off
~ # ls
bin      dev      init     linuxrc  sbin     usr
~ # time
    payload: Any { .. },
    message: Some(
        not yet implemented: Instruction not yet implemented: Unknown,
    ),
    location: Location {
        file: "src/virt.rs",
        line: 312,
        col: 18,
    },
    can_unwind: true,
    force_no_backtrace: false,
} 
[Info  | miralis::debug] Maximal stack usage: 5600 bytes (17.71%)
error: Recipe `run` failed on line 72 with exit code 1
```

**Expected:**
Busybox time command output.

```
~ # time
BusyBox v1.36.0 (2023-03-30 12:28:06 CEST) multi-call binary.

Usage: time [-vpa] [-o FILE] PROG ARGS

Run PROG, display resource usage when it exits

	-v	Verbose
	-p	POSIX output format
	-f FMT	Custom format
	-o FILE	Write result to FILE
	-a	Append (else overwrite)
```

**Reproduce:**

1. checkout commit `ad9f737ac97eab3242cffc5b4f38351398c4e534`
2. set a number of hart >= 1 in the config
3. `just run linux-shell`
4. once in the shell, enter command `time` many times, if you haven't crashed yet

NoeTerrier:
According to specification, `sfence.vma` can be used with registers. (https://riscv-software-src.github.io/riscv-unified-db/html/generic_rv64/insts/sfence.vma.html)

The decoder need then to be modified to handle this kind of `sfence.vma`.

Close #172

### DIFF
--- a/src/arch/metal.rs
+++ b/src/arch/metal.rs
@@ -370,8 +370,29 @@ impl Architecture for MetalArch {
         );
     }
 
-    unsafe fn sfence_vma() {
-        asm!("sfence.vma")
+    unsafe fn sfence_vma(vaddr: Option<usize>, asid: Option<usize>) {
+        match (vaddr, asid) {
+            (None, None) => asm!("sfence.vma"),
+            (None, Some(asid)) => {
+                asm!(
+                    "sfence.vma x0, {asid}",
+                    asid = in(reg) asid,
+                )
+            }
+            (Some(vaddr), None) => {
+                asm!(
+                    "sfence.vma {vaddr}, x0",
+                    vaddr = in(reg) vaddr
+                )
+            }
+            (Some(vaddr), Some(asid)) => {
+                asm!(
+                    "sfence.vma {vaddr}, {asid}",
+                    vaddr = in(reg) vaddr,
+                    asid = in(reg) asid
+                )
+            }
+        }
     }
 
     unsafe fn clear_csr_bits(csr: Csr, bits_mask: usize) {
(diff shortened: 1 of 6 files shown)


src/arch/metal.rs
```diff
@@ -370,8 +370,14 @@ impl Architecture for MetalArch {
         );
     }
 
-    unsafe fn sfence_vma() {
-        asm!("sfence.vma")
+    unsafe fn sfence_vma(rs1: Register, rs2: Register) {
```
NoeTerrier, line 373: I'm not really sure about this part tho, it looks like `usize` could be accepted as register number (same order as x0, x1...) but I would be glad to have confirmation.

CharlyCst: Hmmm, reading the doc it seems the semantic is a bit complicated, because the register `x0` has a special meaning... In the cases where one of the registers is not `x0` we need to load the corresponding value in the virtual register into a physical one and pass the index of that register to the VMA fence. So no it's not really `usizes`, the CPU is indeed going to read the content of the register from my understanding.

CharlyCst (commented):
Thanks for the bugfix on a tricky issue!

It seems the fence is a bit more complicated to emulate than that (see comment), but don't worry I'll implement it based on your code, so enjoy your holiday week :)

PR #176, opened 2024-08-31
Update qemu installation instructions in readme.md

francois141:
The commit explains that to use Miralis, you need to install `qemu-system-riscv64` version 9.0.0 or higher because older versions don't work with it. On Ubuntu, the default `sudo apt-get` install gives you an incompatible version (8.0.0), so you'll need to build QEMU from source. It also shows how to check your QEMU version to make sure you have the right one.


CharlyCst:
I could not reproduce the bug of #154 with an older QEMU version (6.2.0), so closing this for now as it doesn't seem the be the root of the problem.

Issue #182, opened 2024-09-06
Panic on some harts on VisionFive2 board during OpenSBI launch


Wesdcv:
## Issue Description

When running OpenSBI as firmware on our board without spinlocking any harts, we encounter an issue where 1 or 2 harts consistently fail on the first instruction of the OpenSBI binary. The error message received is as follows:

```
[Error | miralis] Panicked at PanicInfo {payload: Any { .. }, 
message: Some(not yet implemented: Other traps are not yet implemented instruction page fault at 40000000), 
location: Location { file: "src/virt.rs", line: 552, col: 21 }, can_unwind: true, force_no_backtrace: false}
```

## Additional Information

- Board model: JH7110 v1.3B
- OpenSBI version: v1.2
- Commit ID: 7b5bfdc219e09e0a8d9e3682ff2cfa0119fe5b82

Wesdcv:
Looks like the faulting harts are `id=4` (consistently) and `id=3` (occasionally).

PR #184, opened 2024-09-10
Fix for issue #182 on the VisionFive2 board

Wesdcv:
The value of physical satp was non-zero on some harts, so firmware in u-mode tried to interpret physical address as virtual on some non-priveleged instructions 
Now we preventively set physical satp to 0 on the board, so that whatever board's spl does with satp before miralis - won't pose an issue
Also moved initial csr setup to architecture init() to clean up code a bit


CharlyCst:
Thanks for the changes! I just fixed a typo and I'll merge once the CI passes :)

CharlyCst (approved):

PR #188, opened 2024-09-12
Spike integration

francois141:


francois141:
@CharlyCst Do you want to have spike in the CI pipeline as well?

CharlyCst:
Awesome, that's going to be super useful, thanks for the changes :)

I made some small changes, in particular removing the feature (because we really want to avoid using `#[cfg]` as much as possible, it is only really necessary when doing architecture specific things which are encapsulated in that `arch` module), and adding a spike config (I think you forgot to include it in the commit^^).
I also removed the `just spike` command, as it is effectively the same as `just run <firmware> config/spike.toml` and I would prefer to avoid adding more commands if the use case can already be covered with the existing ones, but we can always add it back if we need to.

In the future I would like to add Spike in the CI, but not necessarily in the normal `just test` as not everyone might have Spike installed and I really want to keep it simple to get started with Miralis. We will have to think a bit about how to integrate that properly.

PR #189, opened 2024-09-14
Integration of spike

francois141:
@CharlyCst I introduced a small bug in the previous commit. Would it be possible to squash this commit with the last one of master?


CharlyCst:
As a rule we don't modify master (main actually in this case), but we can add a bugfix commit, this kinds of things happen ^^

PR #190, opened 2024-09-15
Wip: Proposition of spike integration in the ci / cd pipeline

francois141:
@CharlyCst  What do you think about this idea to integrate spike in the CI / CD pipeline. We can check if the command is available in the runner and in the case it isn't available the runner simply writes a message to the user and stops the execution. If you think it is a good idea, I can create a clean PR with a compiled version of spike and another configuration file for the last three "external payloads". Feel free to tell me!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,7 @@ jobs:
     - uses: extractions/setup-just@v2 # Install `just`
     - name: Setup Toolchain SPIKE
       run: |
+         git clone https://github.com/riscv-software-src/riscv-isa-sim
          sudo apt-get install device-tree-compiler libboost-regex-dev libboost-system-dev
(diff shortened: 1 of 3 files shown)


CharlyCst:
We could do something like this, for now I would prefer to create a new rule in the justfile `just test-spike`, and run that one also in the CI. It is not the best solution and we will probably remove it later, but we don't want to print 15 "Spike not installed" warning on `just test` either.

I think in the future we will use the runner to orchestrate tests, and it will be able to detect which emulators are present, but it is not the highest priority right now :)

francois141:
@CharlyCst Let me prepare a PR that does that then

francois141:
@CharlyCst What do you think of creating a Docker image where we pre install all dependencies?  I think it will make the CI / CD pipeline much faster.

PR #191, opened 2024-09-17
Add command spike-test

francois141:
This commit adds the command spike-test and allow to run tests with the spike simulator. As some firmware still have a few issues with spike, they are commented out at the moment.


CharlyCst:
Thanks for the changes :)

PR #192, opened 2024-09-17
Integrate cargo deny in the CI / CD pipeline

francois141:


CharlyCst:
I rebased the PR, but turns out installing Cargo Deny in the CI adds 3 more minutes, which is quite a lot...
There seems to be github action that we could use: https://github.com/EmbarkStudios/cargo-deny-action
But then we need to remove the cargo deny installation from `just install`, which will break local tests...

I think we will need a smarter test system in the future (probably as part of the runner), but in the meantime I will simply merge the PR without the bits in the justfile, which means that Cargo Deny will need to be run manually. Its good that we fixed the existing issues though :)

PR #193, opened 2024-09-17
Integrate Clippy (linter) in the CI/CD pipeline

francois141:


francois141:
@CharlyCst This PR is a follow up of the previous one. Clippy seems to be a good extension to cargo fmt and cargo deny

CharlyCst:
It seems that `cargo clippy` only checks the default crate of the workspace (which is the runner), so Miralis itself is not checked it seems.

The whole workspace can be checked with `cargo clippy --workspace --features userspace`. The `userspace` features is necessary to check Miralis for the host target (clippy is a bit lost when checking RISC-V assembly it seems). But there are quite a few more warnings to fix...
Let's keep that around here until we manage to fix most lints :)

francois141:
Good catch! I was a bit surprised to get only warnings in that package. I should have doubled checked. Let me see what I can do for the risc-v assembly and fix the other lints

francois141:
@CharlyCst Ready to review!

CharlyCst:
I rebased the PR (there was a conflict with another one I merged this morning that I fixed), and disabled the deny warning in the justfile to make development easier (the warning are still disabled in the CI by the "RUSTFLAGS" environment variable).

Thanks for pushing the changes, that's super cool :)

Issue #186, opened 2024-09-12
M-mode interrupt virtualization


CharlyCst:
Following  our discussion with @Wesdcv I am opening this issue to track the progress on M-mode interrupt virtualization.

**Context**

Non-delegated interrupts need to be fully virtualized. A non-delegated interrupt will necessarily trap to Miralis because Miralis is the M-mode software. If the corresponding interrupt bit is not set to 0 returning to the firmware or the payload will trap back to Miralis right away, effectively causing an infinite loop on the hart. Disabling the interrupt in `mie` is not an option, because there would be no trap for the next interrupt which would prevent Miralis from injecting the interrupt back into the firmware.

Delegated interrupts do not have this issue, because the OS can handle them without trapping to Miralis. According to the specification all interrupt delegated bits in `mideleg` can be hardwired to 1 except the M-mode interrupt bits. The M-mode interrupts are `MEI`, `MTI`, and `MSI`.
Virtualizing interrupts requires virtualizing the device emitting the interrupt (e.g. the PLIC and CLINT for M-mode interrupts), but we want to limit the virtualization of devices as much as possible. As a result we should force delegation of all non M-mode interrupts.

**Life cycle of virtualized interrupts**

Consider an interrupt MXI that is virtualized.

- The interrupt is not delegated: `mideleg.MXI` is set to 0.
- The interrupt is enabled when it is enabled by the firmware `mie.MXI == vmie.MXI`.
- When Miralis gets the interrupt (it traps with `mip.MXI == 1`) Miralis sets `vmip.MXI` to 1 and deactivates the interrupt. This requires to write to a memory mapped device (e.g. CLINT or PLIC). Note that this device must be virtualized, and thus is not directly accessible by the firmware or the payload. If the interrupt is enabled for the firmware it must be injected, otherwise it must be injected whenever the interrupts become enabled again.
- When the firmware clears the MXI interrupt pending bit by writing to the corresponding (virtualized) memory mapped device Miralis intercepts the write and sets `vmip.MXI` to 0.
- For interrupts that can be received multiple times, such as MIE, Miralis disables the corresponding `mie.MXI` bit if `vmip.MXI` is already 1. This prevent Miralis from trapping again if another interrupts arrives but the previous one has not yet be handled by the firmware. Miralis restores `mie.MXI` to 1 once the firmware disables `vmip.MXI` by writing to the virtualized device.
[...]

PR #199, opened 2024-09-19
Clint interrupts virtualization

Wesdcv:
**Draft PR**: Hopefully partially solves #186 (MEI is yet to implement).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,6 +115,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "702fc72eb24e5a1e48ce58027a675bc24edd52096d5397d4aea7c6dd9eca0bd1"
 
 [[package]]
+name = "clint_interrupt"
+version = "0.1.0"
+dependencies = [
+ "log",
+ "miralis_abi",
+]
+
+[[package]]
+name = "clint_interrupt_multihart"
+version = "0.1.0"
+dependencies = [
+ "log",
+ "miralis_abi",
+]
+
+[[package]]
+name = "clint_interrupt_priority"
+version = "0.1.0"
+dependencies = [
+ "log",
+ "miralis_abi",
+]
+
+[[package]]
 name = "colorchoice"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


Wesdcv:
**Changes Made**  
1. **Filter Added in `set_csr()`**: To prohibit the delegation of the MSI and MTI interrupts. However, MEI should be delegated for S-mode to receive interrupts due to specific PLIC logic. This aspect may need further revision to align with the overall strategy.

2. **MEI Value Update**: Given the unique logic surrounding MEI, its value is currently updated without virtualization. The current approach implemented allows `emulate_jump_trap_handler()` to add pending M-mode interrupts, but not to clear them. While this solution isn't flawless, it functions correctly in both QEMU and Spike under the current non-virtualized MEI conditions.

3. **Pending Interrupt Check**: Each write to the CLINT now checks if any vMXIs (MSI, MTI) should be cleared, ensuring proper interrupt management.

4. **MPIE Virtualization**: I also believe the mstatus.MPIE virtualization used to be implemented without some corner case considerations, but issues only appeared in Spike.

**Questions/Considerations**  
- For CLINT, harts are expected to write to each other's MSIP. However, we might want to ensure that a hart does not attempt to write to another hart's MTIMECMP.
- I left two options in the code for how interrupts should be injected; I would appreciate input on which direction would be best.
- I have left the linux patch in the codebase, mainly because I have no idea what to do with it :) 

I am open to any suggestions or notes for polishing the code, as it may still be a bit raw at this stage. Thank you for your review!

CharlyCst:
Thanks for the work!
I have a deadline for the 18th (on the Tyche project), so I'm a bit busy right now, but I'll come back to work on that and do the review afterward :)

CharlyCst:
I spent some time working on the interrupts, I changed how we handle them in Miralis quite a bit (to add a proper virtualization layer, we don't re-inject the values from the trap-info anymore but we manually emulate the trap when the conditions are satisfied). I ended up using your code during the re-factoring so I will close this PR in favor of #257. I need to test on the board but I think we might finally have a version that works on the board without patching Linux :)

Thanks a lot for all of the work on this, the tests in particular were super helpful and helped me catch a lot of bugs :)
(I marked you as co-author on the commits using you code on the other PR)

PR #200, opened 2024-09-20
Check existence of firmware during the building phase

francois141:


CharlyCst (changes requested):
Overall the new design is cool, but there is just a regression when we pass path to existing binaries on the system (which should be accepted as firmware binaries but that this PR rejects with a non-zero error core).
The correct behavior would be to check if the path exists, and only return an error when it doesn't or is not a file.

CharlyCst (approved):
Thanks for the patch!

I fixed one edge case in where `just run artifacts/opensbi` would not properly load the artifact as an existing binary on the file system, but looks good otherwise.

PR #202, opened 2024-09-21
wip: use logging in the runner & logg when no configuration file is f…

francois141:
…ound


runner/Cargo.toml
```diff
@@ -12,3 +12,5 @@ toml = "0.8.10"
 benchmark_analyzer = { path = "../benchmark_analyzer" }
 config_helpers = { path = "../crates/config_helpers" }
 walkdir = "2"
+env_logger = "0.11.5"
+log = "0.4"
```
CharlyCst, line 16: Replace with `log = { workspace = true }`, this will use the same version as all the other crates, and hopefully save a bit of compile time :)

justfile
```diff
@@ -104,19 +106,19 @@ unit-test:
 
 # Run Miralis
 run firmware=default config=config:
-	cargo run -- run -v --config {{config}} --firmware {{firmware}}
+	cargo run -- -verbose run  --config {{config}} --firmware {{firmware}}
```
CharlyCst, line 109: Typo: `-verbose` => `--verbose`, this causes errors on my side otherwise.

runner/src/artifacts.rs
```diff
@@ -361,20 +361,20 @@ pub fn list_artifacts(args: &ArtifactArgs) {
...
             }
             if let Some(ref repo) = metadata.repo {
-                println!("- [Source repository]({})", repo)
+                log::info!("- [Source repository]({})", repo)
             }
             println!();
         } else {
             // Otherwise simply print the name
-            println!("{}", name)
+            log::info!("{}", name)
```
CharlyCst, line 377: In that one case we actually don't want to use the new logging function,s because it is used to generate the markdown that goes into the website (obviously that won't work anymore with the extra text from `log::info!`). So we need to keep the `println` when `args.markdown == true` in this function.

The page is there for info: https://miralis-firmware.github.io/docs/artifacts

runner/src/main.rs
```diff
@@ -87,6 +87,17 @@ struct ArtifactArgs {
 
 fn main() {
     let args = CliArgs::parse();
+
+    // Set the log level based on the --verbose flag
+    let log_level = if args.verbose {
+        log::LevelFilter::Debug
+    } else {
+        log::LevelFilter::Info
+    };
```
CharlyCst, line 96: That's actually a good idea, I like it :)

CharlyCst (commented):
I like the idea of using the `log` crate here too, thanks for the changes :)

The only one thing I don't like that much is that the output of the `env_logger` is very verbose and not super friendly for a CLI tool. It is great for people running web servers who need precise timestamps and error locations, but for a CLI tool we probably don't need all of that.
What do you think about building our own logger instead? It's not that hard (we do it in Miralis [here](https://github.com/CharlyCst/miralis/blob/main/src/logger.rs), it gives us much more flexibility, and with `std` we can just forward the prints into `println` so it's even simpler). What would be cool is:
- No timestamps, code location, those are just noise for a CLI tool.
- `log::info` is a simple `println`
- `log::warn` prints in yellow bold ([see in Miralis](https://github.com/CharlyCst/miralis/blob/36e7ff195f6b0ea5c5377e2c25cb0bb8266c686e/src/logger.rs#L110) how to do that)
- `log::error` prints in red bold and exits the program with error code 1
- `log::debug` is only printed when `--verbose` is passed, as you did

What do you think of something like this?

PR #205, opened 2024-09-23
Make Miralis panic if we try to set MBE/SBE/UBE bit in mstatus

francois141:
This commit triggers a panic if there is an attempt to set the MBE, SBE, or UBE bits in the mstatus register. This ensures the system crashes intentionally and provides an explicit reason for the crash when these bits are manipulated. The goal is to capture and understand the crash context clearly for debugging purposes.

### DIFF
--- a/src/virt.rs
+++ b/src/virt.rs
@@ -1,6 +1,7 @@
 //! Firmware Virtualisation
 use miralis_core::abi;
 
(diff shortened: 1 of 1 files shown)


CharlyCst (approved):
Awesome, thanks for the change :)

PR #208, opened 2024-09-24
Benchmark payload

francois141:


francois141:
Ready to merge mon capitaine!

config/test/spike-virt-benchmark.toml
```diff
@@ -0,0 +1,23 @@
+# A simple configuration to run on QEMU virt platform
```
CharlyCst, line 1: Small nits, but we should try to update the top-level comment to explain what the config does :)

firmware/tracing_firmware/main.rs
```diff
@@ -0,0 +1,178 @@
...
+        let mcounteren: u32;
+        asm!("csrr {}, mcounteren", out(reg) mcounteren);
+        asm!("csrw mcounteren, {}", in(reg) mcounteren | 1);
+    }
+}
+
+fn main() -> ! {
+    unsafe {
+        asm!("la sp, 0x80700000");
+    }
```
CharlyCst, line 21: You are trying to setup the stack of the firmware here right?
The `setup_binary!` macro already creates a proper stack, so this should not be needed, at least not for the firmware.
The OS will need its own stack though.

firmware/tracing_firmware/main.rs
```diff
@@ -0,0 +1,178 @@
...
+
+
+        os = in(reg) os,
+        mtvec = in(reg) trap,
+        mpp = in(reg) mpp,
+        out("t4") _,
+        );
+    }
+
+    success();
```
CharlyCst, line 49: If the OS is supposed to exit I would suggest to use `failure()` here, as it means we failed to jump into the OS.

firmware/tracing_firmware/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    r#"
+.text
+.align 4
+.global _empty_handler
+_empty_handler:
+    // Skip illegal instruction (pc += 4)
+    csrrw x5, mepc, x5
+    addi x5, x5, 4
+    csrrw x5, mepc, x5
+    // Return back to miralis
```
CharlyCst, line 64: It is not going back to Miralis, but to the OS here.

firmware/tracing_firmware/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    success();
+}
+
+fn trigger_ctx_switch_to_firmware() -> usize {
+    let begin: u64;
+    let end: u64;
+
+    unsafe {
+        // Read the `mcycle` register (assuming 64-bit RISC-V)
+        asm!("csrr {}, cycle", out(reg) begin);
```
CharlyCst, line 124: The only thing we need to be sure here is that reading the cycles doesn't trap :)

You were also talking about taking one big measurement of `NB_REPEATS` traps in a row to double check, do you want to do it here too or do we keep that for future work?

firmware/tracing_firmware/main.rs
```diff
@@ -0,0 +1,178 @@
...
+
+extern "C" {
+    fn _empty_handler();
+}
+
+// —————————————————————————————— Benchmark operating system —————————————————————————————— //
+
+const NB_REPEATS: usize = 1000;
+
+pub fn bubble_sort(arr: &mut [usize; NB_REPEATS]) {
```
CharlyCst, line 77: Nice haha

PR #210, opened 2024-09-24
 Check H & S mode availability at runtime

francois141:


francois141:
@CharlyCst This should avoid a few boring & painful bugs in the future in case we run on a new board or similar

CharlyCst (approved):
Thanks for the changes, looks all good to me :)

PR #212, opened 2024-09-25
Save and restore S registers only if S mode is available

francois141:
Currently we save and restore S-mode registers in the payload in every version of Miralis. If the platform doesn't support S-mode, the code would not work. This commit fixes the issue.


CharlyCst (approved):
Looks all good to me, thanks :)

PR #213, opened 2024-09-26
Add more RAM to qemu

francois141:
@CharlyCst Some juicy memory for our friend qemu....


CharlyCst (approved):
We will have to be careful to ensure that it works on all environments, e.g. even on low memory laptops or in the CI, but so far looks good. We can add a new config option under `[qemu]` if we need it in the future.

PR #214, opened 2024-09-26
Enable hypervisor mode if qemu-system-riscv64 is 6

francois141:
In the version 6 of qemu-system-riscv64, we don't the the hypervisor extension enabled by default. This is problematic when we want to run qemu in qemu. This commits adds the required flag.


francois141:
@CharlyCst You can already review. Tomorrow I test it on my lab machine

runner/src/run.rs
```diff
@@ -85,6 +99,11 @@ fn launch_qemu(args: &RunArgs, miralis: PathBuf, firmware: PathBuf) -> ExitCode
     if let Some(cpu) = &cfg.qemu.cpu {
         qemu_cmd.arg("-cpu").arg(cpu);
     }
+
+    if is_qemu_version_6() {
```
CharlyCst, line 103: Let's add a comment above there, to explain why we need to pass an additional flag for some QEMU version.

In general this kind of stuff can not be inferred from the code, so it is super valuable to document why we add those kinds of edge-cases.

runner/src/run.rs
```diff
@@ -75,6 +75,20 @@ fn get_config(args: &RunArgs) -> Config {
     cfg
 }
 
+fn is_qemu_version_6() -> bool {
```
CharlyCst, line 78: Maybe let's do a slightly more general purpose function `get_qemu_version` which returns the version string.
We might need it for other purposes in the future (e.g. warn that some features might be missing, or simply report the version to the user in case of error, etc...).

runner/src/run.rs
```diff
@@ -75,6 +75,20 @@ fn get_config(args: &RunArgs) -> Config {
...
+
+    let output = Command::new("qemu-system-riscv64")
+        .arg("--version")
+        .output()
+        .expect("Failed to execute command");
```
CharlyCst, line 83: Be a bit more specific here, it might not be clear for the user what command failed.

Something like "Failed to execute `qemu-system-riscv64 --version`" would be more helpful to diagnose an issue.

runner/src/run.rs
```diff
@@ -85,6 +99,11 @@ fn launch_qemu(args: &RunArgs, miralis: PathBuf, firmware: PathBuf) -> ExitCode
     if let Some(cpu) = &cfg.qemu.cpu {
         qemu_cmd.arg("-cpu").arg(cpu);
     }
+
+    if is_qemu_version_6() {
+        qemu_cmd.arg("-cpu").arg("-cpu rv64,h=true");
```
CharlyCst, line 104: Isn't this supposed to be `qemu_cmd.arg("-cpu").arg("rv64,h=true");` ? You repeated `-cpu` twice.

CharlyCst, line 104: I would also add a `log::debug!` here to tell the user we enable the `h` extension. This might cause some unexpected and hard to debug behavior (say if in the future we add some conflicting CPU option, or if the user sets `h=false` in its config which is already possible today).

CharlyCst (changes requested):
Thanks for the change :)

I added a few comments, mostly to future proof the code and make it easier to trace bugs in the future.

CharlyCst (approved):
I changed a bit the code the extract the version number :)

I realized that we can already have this behavior by changing the configuration (there is a field to specify the QEMU cpu features). If you create a default configuration `./config.toml` the runner will take it into account even if you don't specify the path, so it should not be too annoying ^^

But let's merge this anyway :)

CharlyCst:
Hmmm, seems like the CI doesn't like the flag:
`qemu-system-riscv64: can't apply global rv64-riscv-cpu.h=true: Property 'rv64-riscv-cpu.h' not found`

It might be even older?

francois141:
@CharlyCst Ready to review, now it works. The flag was wrong

CharlyCst:
Should we re-open the PR or keep it closed?

PR #217, opened 2024-09-28
Bugfix: correction of emulation in HStatus

francois141:
Currently the hstatus register isn't emulated perfectry. This commit fixes the issue.


CharlyCst (approved):
Cool, looks all good to me :)

PR #220, opened 2024-09-29
Bugfix: set TVM and TSR fields of Mstatus read-only only when S-mode …

francois141:
…isn't available

The code assumed that we didn't have S-mode potentially available, which is a wrong assumption. This commit fixes the issue.


CharlyCst:
Thanks for the bugfix.

This is typically the kind of things we hope to catch with the formal verification of privileged instruction emulation. So hopefully eventually we will have a principled way to fix all such bugs.

PR #221, opened 2024-09-29
Bugfix: properly display debug message in runner

CharlyCst:
The runner was recently updated with a new logging facility, but turns out it was not displaying debug-level messages even with the `--verbose` flag.
This commit fixes the issue, and cleans the output a bit more by removing another non-useful log. The runner now properly displays the QEMU command when invoked with verbose.

### DIFF
--- a/runner/src/logger.rs
+++ b/runner/src/logger.rs
@@ -47,10 +47,9 @@ impl log::Log for RunnerLogger {
                 Level::Warn => {
                     println!("\x1b[33m{}\x1b[0m", record.args());
                 }
-                Level::Info => {
+                Level::Info | Level::Debug | Level::Trace => {
                     println!("{}", record.args());
                 }
-                _ => {}
             }
         }
     }
--- a/runner/src/artifacts.rs
+++ b/runner/src/artifacts.rs
@@ -226,8 +226,6 @@ pub fn build_target(target: Target, cfg: &Config) -> PathBuf {
             }
         }
     };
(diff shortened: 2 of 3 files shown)


runner/src/logger.rs
francois141, line 1: @CharlyCst I think we can also add trace here. It might save us a few stupid bugs in the future :)

francois141: Good catch and my bad!

CharlyCst:
Wow, we are hitting a flaky unit test, interesting.

I could reproduce it once locally with `just test`, so it's not specific to the CI:

```
test virt::tests::switch_context_mpp ... FAILED

failures:

---- virt::tests::switch_context_mpp stdout ----
thread 'virt::tests::switch_context_mpp' panicked at src/virt.rs:1540:9:
assertion `left == right` failed: Mstatus.MPP must be set to M mode (coming from Miralis)
  left: 0
 right: 6144
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    virt::tests::switch_context_mpp
```

It seems I can't reproduce it anymore... If this happens again we should investigate.

CharlyCst:
Re-running the pipeline made the test succeed.
It seems like we have a low-probability flaky test, which is not great... Let's try to collect more info next time we run into it.

runner/src/logger.rs
CharlyCst: I added the trace level too :)

PR #224, opened 2024-10-01
Implementation of ProtectPayload

francois141:
This commit introduces the ProtectPayload policy, enabling the Payload to make an ecall to the security monitor during execution, thereby safeguarding itself from potential firmware interference.


src/policy/protectpayload.rs
```diff
@@ -0,0 +1,104 @@
+//! The default policy module, which enforces no policy.
```
CharlyCst, line 1: Change the description here to explain what this policy does.

In general the rust module-level documentation is formatted as a title, followed by a blank line, and a descriptive paragraph. Like this:

```rs
//! Example of Module-Level Documentation
//!
//! The module-level documentation start with `//!` instead of `//!`.
//! Is is useful to write some, they are displayed both in the auto-generated docs with
//! `cargo doc --open` and by your IDE.

pub my_code_goes_here;
```

src/policy/protectpayload.rs
```diff
@@ -0,0 +1,104 @@
...
+use miralis_core::abi;
+
+use crate::arch::pmp::{build_napot, pmpcfg};
+use crate::arch::{Arch, Architecture, Register};
+use crate::host::MiralisContext;
+use crate::policy::{PolicyHookResult, PolicyModule};
+use crate::virt::{RegisterContextGetter, VirtContext};
+
+/// The default policy module, which doesn't enforce any isolation between the firmware and the
+/// rest of the system.
```
CharlyCst, line 12: Change the description here too.

src/policy/protectpayload.rs
```diff
@@ -0,0 +1,104 @@
...
+            return;
+        }
+
+        // Step 1: Clear general purpose registers
+        for i in 0..32 {
+            self.general_register[i] = ctx.regs[i];
+            // We don't clear ecall registers
+            if !(10..18).contains(&i) {
+                ctx.regs[i] = 0;
+            }
```
CharlyCst, line 68: As a next step we will clear them if the trap is not caused by an `ecall`.

src/policy/protectpayload.rs
```diff
@@ -0,0 +1,104 @@
...
+        if !self.protected {
+            return;
+        }
+
+        // Step 1: Restore general purpose registers
+        for i in 0..32 {
+            // 10 & 11 are return registers
+            if (10..12).contains(&i) {
+                continue;
+            }
```
CharlyCst, line 82: Those you be kept only if this is a return from an `ecall`, which means that we need to remember the cause of the trap somehow.

src/policy/protectpayload.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl ProtectPayloadPolicy {
+    fn lock(&mut self, mctx: &mut MiralisContext) {
+        // First set pmp entry protection
+        mctx.pmp.set(
+            mctx.devices.len() + 2,
+            build_napot(0x80400000, 0x8000).unwrap(),
+            pmpcfg::NAPOT,
+        );
```
CharlyCst, line 95: We will need a clean API to use the set of PMPs reserved for the Policy Module at some point, but I guess we can think bout that when we improve the PMP module.

src/policy/protectpayload.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+}
+
+impl ProtectPayloadPolicy {
+    fn lock(&mut self, mctx: &mut MiralisContext) {
+        // First set pmp entry protection
+        mctx.pmp.set(
+            mctx.devices.len() + 2,
+            build_napot(0x80400000, 0x8000).unwrap(),
```
CharlyCst, line 93: Here you a protecting a range of size 0x8000, it's cool for testing, but we should probably take the size as an argument in the future (so that Linux can protect its whole memory for instance).

src/virt.rs
```diff
@@ -44,10 +44,8 @@ pub struct VirtContext {
     pub(crate) mode: Mode,
     /// Number of virtual PMPs
     pub(crate) nb_pmp: usize,
-    /// RISC-V Hypervisor extension available
-    pub(crate) has_h_mode: bool,
-    /// RISC-V Supervisor extension available
-    pub(crate) has_s_mode: bool,
+    /// Availables RISC-V extensions
+    pub(crate) extensions: ExtensionsCapability,
```
CharlyCst, line 48: Do we need to have it both in `mctx` and the `VirtContext`? I don't remember... If that's the case fine :)

src/virt.rs
```diff
@@ -562,11 +558,19 @@ impl VirtContext {
     }
 
     /// Handle the trap coming from the firmware
-    pub fn handle_firmware_trap(&mut self, mctx: &mut MiralisContext) {
+    pub fn handle_firmware_trap(&mut self, mctx: &mut MiralisContext, policy: &mut Policy) {
         let cause = self.trap_info.get_cause();
         match cause {
-            MCause::EcallFromSMode if Policy::ecall_from_firmware(self).overwrites() => {
+            MCause::EcallFromSMode if policy.ecall_from_payload(mctx, self).overwrites() => {
```
CharlyCst, line 564: Is it `ecall_from_payload` or for firmware? I guess here we should never get an `EcallFromSMode` anyway as we come from the firmware which is in U-mode.

src/virt.rs
```diff
@@ -562,11 +558,19 @@ impl VirtContext {
...
                 // Nothing to do, the Policy module handles those ecalls
+                log::info!("Catching E-call from payload in the policy module");
+            }
+            MCause::EcallFromUMode
+                if self.get(Register::X17) == abi::MIRALIS_EID
+                    && policy.ecall_from_firmware(mctx, self).overwrites() =>
```
CharlyCst, line 570: I thin the policy and the Miralis `ecall`s should be two distinct things. So each policy module can have its own `EID`.

src/virt.rs
```diff
@@ -603,7 +607,7 @@ impl VirtContext {
                     // TODO: make sure virtual address does not get around PMP protection
                     let instr = unsafe { Arch::get_raw_faulting_instr(&self.trap_info) };
                     let instr = mctx.decode(instr);
-                    log::trace!(
+                    log::info!(
```
CharlyCst, line 610: let's keep that a trace, or maybe a `debug`.

src/virt.rs
```diff
@@ -620,7 +624,7 @@ impl VirtContext {
                 }
             }
             MCause::InstrAccessFault => {
-                log::trace!("Instruction access fault: {:x?}", self.trap_info);
+                log::info!("Instruction access fault: {:x?}", self.trap_info);
```
CharlyCst, line 627: Same comment

CharlyCst (commented):
Great job on this first iteration, I made some comments so we can start refining the code

src/main.rs
```diff
@@ -110,7 +110,8 @@ pub(crate) extern "C" fn main(_hart_id: usize, device_tree_blob_addr: usize) ->
         }
 
         // Add an inactive 0 entry so that the next PMP sees 0 with TOR configuration
-        let inactive_index = devices.len() + 2;
+        let inactive_index = devices.len() + 3;
+        mctx.pmp.set(inactive_index - 1, 0, pmpcfg::INACTIVE);
         mctx.pmp.set(inactive_index, 0, pmpcfg::INACTIVE);
```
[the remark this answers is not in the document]
francois141: I see. Let's do it in another iteration I would say

src/policy/default.rs
```diff
@@ -1,22 +1,47 @@
...
+        &mut self,
+        _mctx: &mut MiralisContext,
+        ctx: &mut VirtContext,
+    ) -> PolicyHookResult {
+        if ctx.get(Register::X16) == abi::MIRALIS_LOCK_FID {
+            ctx.pc += 4;
+            PolicyHookResult::Overwrite
+        } else {
+            PolicyHookResult::Ignore
+        }
```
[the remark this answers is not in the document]
francois141: True I added for debugging. We can remove it

src/policy/protectpayload.rs
francois141: @CharlyCst Can we do this in another iteration? I plan to open an issue with the remaining steps and go step by step :=)

src/policy/protectpayload.rs
francois141: I opened an issue for that :-)

src/virt.rs
francois141: This make it easier :-)

src/main.rs
```diff
@@ -138,7 +139,7 @@ pub(crate) extern "C" fn main(_hart_id: usize, device_tree_blob_addr: usize) ->
     }
 
     // Initialize the virtual context and configure architecture
-    let mut ctx = VirtContext::new(hart_id, nb_virt_pmp, mctx.hw.has_h_mode, mctx.hw.has_s_mode);
+    let mut ctx = VirtContext::new(hart_id, nb_virt_pmp, mctx.hw.extensions.clone());
```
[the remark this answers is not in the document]
francois141: It needs

src/virt.rs
francois141: True, mistake from my side

crates/abi/src/lib.rs
```diff
@@ -52,6 +52,11 @@ pub fn miralis_end_benchmark() -> ! {
     }
 }
 
+/// Ask Miralis to lock the payload
+pub fn lock_payload() {
+    unsafe { miralis_ecall(abi_protect_payload::MIRALIS_PROTECT_PAYLOAD_LOCK_FID).ok() };
+}
+
```
CharlyCst, line 59: Since we created a new EID for the policy we can't use `miralis_ecall` (because it uses the Miralis EID). Let's create a new function `policy_ecall` which is similar to `miralis_ecall` but also takes the policy EID as a parameter, and use it for the `lock_payload` ^^

firmware/protect_payload/main.rs
```diff
@@ -0,0 +1,69 @@
+//! Protect payload test
+//!
+//! This test firmware checks that Miralis is properly protected, or in other words that the
+//! firmware is properly sandboxed.
```
CharlyCst, line 4: A small nit: here it we don't check that Miralis is protected, we check that the protect payload policy works as expected. We should probably at least mention that the test is intended to be run with the Protect Payload policy (which is not enabled by default, so might be surprising to someone who tries to run the test with `just run protect_payload`)

src/policy/mod.rs
```diff
@@ -4,13 +4,16 @@
 
 use config_select::select_env;
 
+use crate::host::MiralisContext;
 use crate::virt::VirtContext;
 
 mod default;
 mod keystone;
+mod protectpayload;
```
CharlyCst, line 12: Small nit, but for naming consistency let's use `protect_payload` with an underscore (`_`)

CharlyCst (commented):
Just a few small nits/clarification, and I think we're good to go for the first merge! 🥳

firmware/protect_payload/main.rs
CharlyCst: Looks like you rolled back on this file, it's the old text and old code

firmware/protect_payload/main.rs
```diff
@@ -0,0 +1,147 @@
...
+    // Make sure r/w to Payload leads to an error
+    try_rw_protected_payload();
+
+    // We can exit with success
+    success();
+}
+
+fn try_rw_miralis() {
+    let mut mcause: usize;
+    let x: usize = 0x800020000;
```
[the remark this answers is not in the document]
CharlyCst: Similar, looks like you rolled back: the name is reverted to `x`

firmware/protect_payload/main.rs
```diff
@@ -0,0 +1,147 @@
...
+    );
+}
+
+fn try_rw_protected_payload() {
+    unsafe {
+        asm!("csrw mcause, zero");
+    }
+
+    let mut mcause: usize;
+    let x: usize = 0x80400000;
```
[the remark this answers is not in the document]
CharlyCst: Same thing here

CharlyCst (commented):
I think you just rolled back some changes on a file, but beside that looks all good to me :)

CharlyCst (approved):
Awesome, all good! Good job on getting this first part done :)

PR #227, opened 2024-10-02
Implement Keystone Policy

FredKhayat:


francois141:
@CharlyCst @FredKhayat I had to modify a bit the interface of the policy module for my own policy. I don't think it should be a big deal for the keystone policy though....

FredKhayat:
> @CharlyCst @FredKhayat I had to modify a bit the interface of the policy module for my own policy. I don't think it should be a big deal for the keystone policy though....

Yeah thanks, I actually needed some of the changes to continue working on the keystone policy

PR #231, opened 2024-10-03
Refactoring PMP module

francois141:
Currently the initialisation of the PMP in main.rs is hardcoded and isn't scalable over time. This commit solves the issue.


CharlyCst (approved):
Looks good, thanks :)

PR #232, opened 2024-10-03
Create minimal test infrastructure for testing protect payload.

francois141:
We now aim to identify the specific requirements for the protect payload policy and explore the various scenarios that might arise. This commit establishes the foundation of a test framework to ensure that Miralis consistently enforces the necessary invariants.


payload/hello_world/main.rs
```diff
@@ -3,9 +3,9 @@
 #![feature(start)]
 // ———————————————————————————————— Guest OS ———————————————————————————————— //
 
-use miralis_abi::{log, setup_binary, success};
+use miralis_abi::{log, setup_binary, setup_payload, success};
```
CharlyCst, line 6: So you need to import both `setup_payload` and `setup_binary` there because `setup_payload` is a macro and expects a macro called `setup_binary` in scope, that's not great...

On second thought I think it is better to stick with `setup_binary` only. Granted the name is not great, but it is easier to maintain a single macro (for instance we don't have a doc comment on `setup_payload` but we do in `setup_macro`). I don't think keeping two distinct macros brings enough benefits on the table, let's keep things simple.

firmware/test_protect_payload_firmware/main.rs
```diff
@@ -0,0 +1,33 @@
+#![no_std]
```
CharlyCst, line 1: It would be good to have a few doc comments here, those tests are going to be non-trivial, so describing what is happening (e.g. once the workflow you want to test is clearer we should describe it in a doc comment). This is true for the other firmware/payloads too.

CharlyCst (commented):
So there is going to be two tests for the protect payload policy if I understand correctly:
One with a single firmware, and one with a firmware + a payload. Are they testing different things?

==== EDIT ====

For the moment yes and this is redundant. When I start implementing the tests for the protect payload policy, I will remove the first test and keep only a single one.

firmware/test_protect_payload_firmware/main.rs
francois141: classique comment :)

francois141:
@CharlyCst Ready to review again. As a well organized student, I forgot to take with me my txt file yesterday. Therefore I am writing on extending the policy and this afternoon when I come, I will continue working on the doc

CharlyCst (approved):
Awesome, looks all good :)

PR #235, opened 2024-10-04
Bugfix: deactivate PMP before jumping in S-mode in protect_payload.rs

francois141:
This commit achieves two objectives. First, it deactivates the PMP entry when transitioning into the S-mode payload. Second, it introduces several tests for the protect payload policy.


src/policy/protect_payload.rs
```diff
@@ -57,6 +57,7 @@ impl PolicyModule for ProtectPayloadPolicy {
         }
 
         if ctx.get(Register::X16) == abi_protect_payload::MIRALIS_PROTECT_PAYLOAD_LOCK_FID {
+            log::info!("Locking payload from firmware");
```
CharlyCst, line 60: We probably don't want those to be `info` level in the future

CharlyCst (approved):
Cool, thanks for the changes :)

PR #236, opened 2024-10-04
Rebased commit explicit forwarding

francois141:


CharlyCst:
It seems there are some conflicts with another PR you pushed (and that I merged, but I assume I should have merge the two commits there instead...)

You should cherry-pick the changes from the second commit to put them on a new branch, to avoid duplicating the first commit.
If you are still iterating over this benchmark maybe you should iterate on a branch for a bit and merge only later when you are satisfied, rather than creating lots of commits with experimental changes.

francois141:
Weird, I did this. Let me have a look tomorrow, end the benchmark and create a new pr. I am almost done with this benchmark in reality.

PR #237, opened 2024-10-06
Add payload hash function

francois141:
This commits introduces a function that allows to hash the binary content of the payload. This will allow us in the future to check if the payload was loaded correctly and without malicious actions by the firmware.

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -17,6 +17,8 @@ miralis_core = { path = "../crates/core" }
 miralis_abi = { path = "../crates/abi" }
 config_helpers = { path = "../crates/config_helpers" }
 config_select = { path = "../crates/config_select/" }
+# This import is only used in the protect payload policy
+tiny-keccak = { version = "2.0.0", features = ["sha3"] }
 
 [features]
 # When running on host architecture as a userspace application, such as when
--- a/src/policy/protect_payload.rs
+++ b/src/policy/protect_payload.rs
@@ -1,15 +1,32 @@
 //! The protect payload policy, it protects the payload.
 
+use core::slice;
+use core::sync::atomic::{AtomicBool, Ordering};
+
 use miralis_core::abi_protect_payload;
+use tiny_keccak::{Hasher, Sha3};
 
 use crate::arch::pmp::pmpcfg;
 use crate::arch::pmp::pmplayout::POLICY_OFFSET;
 use crate::arch::MCause::EcallFromSMode;
 use crate::arch::{MCause, Register};
+use crate::config::PAYLOAD_HASH_SIZE;
 use crate::host::MiralisContext;
 use crate::platform::{Plat, Platform};
 use crate::policy::{PolicyHookResult, PolicyModule};
 use crate::virt::{RegisterContextGetter, VirtContext};
+
+const LINUX_LOCK_PAYLOAD_HASH: [u8; 32] = [
+    241, 90, 158, 184, 200, 210, 145, 178, 30, 80, 200, 161, 56, 120, 75, 241, 68, 38, 21, 2, 248,
+    112, 128, 155, 31, 240, 37, 94, 203, 66, 243, 167,
+];
+
+const TEST_POLICY_PAYLOAD: [u8; 32] = [
+    138, 39, 74, 50, 74, 113, 151, 180, 78, 91, 92, 25, 132, 84, 192, 119, 38, 36, 19, 147, 193,
+    166, 149, 149, 158, 179, 237, 5, 158, 54, 245, 69,
+];
+
+static FIRST_JUMP: AtomicBool = AtomicBool::new(true);
 
 /// The protect payload policy module, which allow the payload to protect himself from the firmware at some point in time and enfore a boundary between the two components.
 pub struct ProtectPayloadPolicy {
@@ -17,7 +34,6 @@ pub struct ProtectPayloadPolicy {
     general_register: [usize; 32],
     rules: [ForwardingRule; ForwardingRule::NB_RULES],
     last_cause: MCause,
-    first_jump: bool,
 }
 
 impl PolicyModule for ProtectPayloadPolicy {
@@ -29,7 +45,6 @@ impl PolicyModule for ProtectPayloadPolicy {
             // It is important to let the first mode be EcallFromSMode as the firmware passes some information to the OS.
             // Setting this last_cause allows to pass the arguments during the first call.
             last_cause: MCause::EcallFromSMode,
-            first_jump: true,
         }
     }
     fn name() -> &'static str {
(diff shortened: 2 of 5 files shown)


CharlyCst (commented):
There are some details to iron-out on the hashing side, in particular hashing the memory is not sufficient if the state of the registers is not also known.

There are multiple ways of doing so, we could either hash the registers too, or include them (and sign them) in the attestation.

src/policy/protect_payload.rs
```diff
@@ -212,3 +215,22 @@ impl ProtectPayloadPolicy {
...
+
+#[allow(dead_code)]
+fn hash_memory_range(start: usize, end: usize) -> [u8; 32] {
+    assert!(start <= end, "Invalid memory range for payload hashing");
+
+    unsafe {
+        let data = slice::from_raw_parts(start as *const u8, end - start);
+
+        let mut hasher = Sha3::v256();
+        hasher.update(data);
```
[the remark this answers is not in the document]
francois141: I don't think we need to hash the registers, because between the firmware and the palyoad we have the policy that can filter them. The hash is here because we need to make sure the payload is correctly loaded. Do you see what I mean?

src/policy/protect_payload.rs
```diff
@@ -212,3 +215,22 @@ impl ProtectPayloadPolicy {
...
 }
+
+// ———————————————————————————————— Hash primitive ———————————————————————————————— //
+
+#[allow(dead_code)]
+fn hash_memory_range(start: usize, end: usize) -> [u8; 32] {
+    assert!(start <= end, "Invalid memory range for payload hashing");
+
+    unsafe {
+        let data = slice::from_raw_parts(start as *const u8, end - start);
```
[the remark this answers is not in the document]
francois141: It is readonly. Which consequences could we have?

CharlyCst (commented):
The commit looks good, but let's not merge this until the feature is complete (we can add more commit to the branch).

In particular hashing is only useful if the payload can get the hash. So the next steps there would be to add a new `ecall` to the protect payload policy to retrieve the hash.
Unfortunately it's a bit tricky, because the SBI call ABI specification says we can only use a single register to return a value (`a1` see chapter 3 [here](https://tools.cloudbear.ru/docs/riscv-sbi-2.0-20231006.pdf)), so that's 64 bits but we would need at least 256.

There are multiple options:
- We break the ABI and use more registers to return values. We don't want to do that because then the protect payload SBI becomes incompatible with all the other SBI, so we can't re-use the existing SBI infrastructure.
- We return 64 bits at a time, and we expect the caller to do multiple calls to retrieve the full hash. We would need one of the parameters to be an index, but otherwise the implementation is straight forward. The downside is that if we grow the attestation over time it will become more and more costly to retrieve all of it.
- We take a pointer from the host and write the attestation there. The safest way of doing so is to use the `mstatus.MPRV` bit to ensure Miralis does not end up being a [confused deputy](https://en.wikipedia.org/wiki/Confused_deputy_problem). That requires some delicate assembly that muse be carefully crafted to not introduce any vulnerability.

I would suggest to start with option 2 personally :)

CharlyCst:
When responding to a comment please do not edit the original one, simply use the "comment" button. You can quote comments in markdown like that:

> === Edit @francois141 ===
> 
> My implementation checks directly in the security monitor. What do you think about that?

So avoid doing the thing above, and isntead use markdown quotes (lines starting with `> `) and pinging people if needed.

I will edit my comment to remove your modifications.

src/policy/protect_payload.rs
```diff
@@ -117,6 +135,25 @@ impl PolicyModule for ProtectPayloadPolicy {
         // Step 3: Unlock memory
         mctx.pmp.set_inactive(POLICY_OFFSET, 0x80400000);
```
CharlyCst, line 136: I a not sure how I feel about hard-coded constants there, keep in mind that we will need to get it to work on the board eventually, so those addresses won't work.

There is a `TARGET_PAYLOAD_ADDRESS` constant, would that work instead?

src/policy/protect_payload.rs
```diff
@@ -1,21 +1,38 @@
...
 
+const LINUX_LOCK_PAYLOAD_HASH: [u8; 32] = [
+    241, 90, 158, 184, 200, 210, 145, 178, 30, 80, 200, 161, 56, 120, 75, 241, 68, 38, 21, 2, 248,
+    112, 128, 155, 31, 240, 37, 94, 203, 66, 243, 167,
+];
+
+const TEST_POLICY_PAYLOAD: [u8; 32] = [
+    140, 165, 21, 156, 234, 234, 86, 210, 171, 224, 126, 232, 161, 252, 107, 57, 157, 200, 184,
+    100, 29, 28, 164, 226, 250, 171, 137, 218, 97, 135, 196, 86,
+];
```
CharlyCst, line 27: Keep in mind that we will need to check the hash of different kernels, for instance the one on the board is a issued from a fork maintained by StarFive, so we need support for at least two kernel images and the test payload.

src/Cargo.toml
```diff
@@ -17,6 +17,9 @@ miralis_core = { path = "../crates/core" }
 miralis_abi = { path = "../crates/abi" }
 config_helpers = { path = "../crates/config_helpers" }
 config_select = { path = "../crates/config_select/" }
+# This import is only used in the protect payload policy
+tiny-keccak = { version = "2.0.0", features = ["sha3"] }
+fdt-rs = { version = "0.4", default-features = false, features = [] }
```
CharlyCst, line 22: Maybe I missed it, but I don't see it used in this commit?

src/policy/protect_payload.rs
```diff
@@ -192,3 +229,31 @@ impl ForwardingRule {
...
+
+    unsafe {
+        let mut hasher = Sha3::v256();
+
+        let payload_content =
+            slice::from_raw_parts(payload_start as *const u8, payload_end - payload_start);
+        hasher.update(payload_content);
+
+        let pc_start = slice::from_raw_parts(pc_start as *const u8, 4);
+        hasher.update(pc_start);
```
CharlyCst, line 252: We want to hash the *value* of the PC, not the *data pointed by* the PC. If you hash the data pointed by there is very little guarantees, as there could be lots of entry points starting with the same instruction.

If you need an array of bytes for the hasher, you can use `pc_start.to_le_bytes()`.

src/policy/protect_payload.rs
```diff
@@ -192,3 +229,31 @@ impl ForwardingRule {
         self
     }
 }
+
+// ———————————————————————————————— Hash primitive ———————————————————————————————— //
+
+fn hash_payload(size_to_hash: usize, pc_start: usize) -> [u8; 32] {
+    let payload_start: usize = 0x80400000;
+    let payload_end: usize = 0x80400000 + size_to_hash;
```
CharlyCst, line 237: Unfortunately this is not going to work on the board, the addresses are different.

The configuration file already has a PAYLAOD_START_ADDRESS field, could you use that instead? We have a `TARGET_PAYLOAD_ADDRESS` constant defined in `config.rs`.

CharlyCst (commented):
Here are a few comments, the most critical issue is the absolute addresses, those won't work on the board.

francois141:
@CharlyCst I am aware regarding the absolute addresses. I want to tackle this in a new commit as we also hardcode the 0x80400000 at other places. I want to do a single commit that make it work for the board. Does that sounds good to you?

src/policy/protect_payload.rs
francois141: This is not part of the commit, I guess we should fix this in another one

src/policy/protect_payload.rs
francois141: I would suggest to fix this in another commit properly with the rest of the protect payload policy

francois141:
Ready for review!

CharlyCst (approved):

Modulo rebasing the PR is ready.

There is one caveat though, it is that this will break the Protect Payload policy on the board (the hash won't be the same). Are we fine with that for now?

francois141:
For me this is not an issue because we will receive the message on the board if we forget and I will just add the hash  (which I don't have now)

CharlyCst:
@francois141 I rebased the PR (please update your locate to the remote branch to pick-up the changes)

It turns out exact hashes are a problem with the CI: the build does not seem to be 100% deterministic so I am getting a different hash than yours and the CI yet another one. Currently this PR breaks the tests.

As a start I would suggest simply logging the hashes instead of crashing to at least keep the tests passing. What do you think?

francois141:
This sounds good to me. We can as a next step implement a memory clearing option for the CI / CD pipeline

PR #239, opened 2024-10-06
Creation of explicit forwarding rules in the protect_payload.rs

francois141:
This commit creates a small framework similar to IPRules that allows to explicitly forwards registers from the payload to the firmware and in the opposite direction. Last but not least, it creates and test a forwarding rule for ecall from S mode.

### DIFF
--- a/payload/test_protect_payload_payload/main.rs
+++ b/payload/test_protect_payload_payload/main.rs
@@ -60,22 +60,6 @@ macro_rules! save_registers {
 
     };
 }
-
-macro_rules! test_same_registers_after {
-    ($asm_code:expr) => {{
-        let mut regs_before = Regs::new();
-        let mut regs_after = Regs::new();
-
-        unsafe {
-            save_registers!(regs_before);
-            asm!($asm_code);
-            save_registers!(regs_after);
-        }
-
-        regs_before == regs_after
-    }};
-}
-
 // ———————————————————————————————— Guest OS ———————————————————————————————— //
 
 use core::arch::asm;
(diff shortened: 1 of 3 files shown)


src/policy/protect_payload.rs
```diff
@@ -212,3 +216,87 @@ impl ProtectPayloadPolicy {
...
+            .allow_register_in(Register::X15)
+            .allow_register_in(Register::X16)
+            .allow_register_in(Register::X17)
+            .allow_register_out(Register::X10)
+            .allow_register_out(Register::X11);
+
+        rules
+    }
+
+    #[allow(unused)]
```
CharlyCst, line 261: It is used so no need for this annotation.

src/policy/protect_payload.rs
```diff
@@ -212,3 +216,87 @@ impl ProtectPayloadPolicy {
...
+    fn allow_register_out(&mut self, reg: Register) -> &mut Self {
+        self.allow_out[reg as usize] = true;
+        self
+    }
+
+    #[allow(unused)]
+    fn block_register_out(&mut self, reg: Register) -> &mut Self {
+        self.allow_out[reg as usize] = false;
+        self
+    }
```
CharlyCst, line 301: In general, let's remove the `#[allow]` annotation and only keep the one we actually use.

The problem with those `#[allow]` annotation is that we always end-up forgetting to remove then, and then we end-up with a lot of dead-code that the compile doesn't warn about...

CharlyCst (commented):
Looks cool, I mostly have some cleanup remarks :)

payload/test_protect_payload_payload/main.rs
```diff
@@ -6,6 +6,76 @@
...
+impl Regs {
+    pub fn new() -> Self {
+        Regs { registers: [0; 32] }
+    }
+}
+
+macro_rules! save_registers {
+    ($regs:expr) => {
+            asm!(
+                "sd x0, 0*8({0})",
```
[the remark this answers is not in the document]
francois141: True :)

PR #240, opened 2024-10-08
Bugfix: replace mctx.virt_pmp_offset by mctx.pmp_virt_pmp

francois141:
And the mistery of the PMP problem has been solved


francois141:
@CharlyCst I know the protect payload policy test is broken now. In the commit with explicit forwarding rules I will fix everything.

CharlyCst (approved):
Looks all good, thanks for the fix :)

PR #244, opened 2024-10-17
Start Implementing Keystone Policy

FredKhayat:


config/qemu-keystone.toml
```diff
@@ -16,3 +16,19 @@ enable = false
...
+start_address = 0x80000000
+stack_size = 0x8000
+
+[target.firmware]
+name="opensbi-jump"
+profile = "dev"
+start_address = 0x80200000
+
+[target.payload]
+name = "./images/Image"
```
CharlyCst, line 31: I suppose this is an internal path? We have a concept of artifacts that we can use for that. You can learn more about it [here](https://miralis-firmware.github.io/docs/tooling#test-artifacts) and see the existing list [here](https://miralis-firmware.github.io/docs/artifacts).

Eventually we will want to setup a repository to build all the required artifacts for Keystone. You can see an example [here]. The idea is simply to build every binary in the CI and expose them in the release, then we can update the [`artifacts.toml`](https://github.com/CharlyCst/miralis/blob/8d767b6a3f1605d33d0a38d545b431720fc3c221/misc/artifacts.toml) file to point to the new binaries.

If you feel stuck on porting Keystone to Miralis you can also look into that eventually :)

src/policy/keystone.rs
```diff
@@ -3,24 +3,176 @@
 //! This policy module enforces the Keystone policies, i.e. it enables the creation of user-level
 //! enclaves by leveraging PMP for memory isolation.
 
+#![allow(unused)]
```
CharlyCst, line 6: In general we don't want to add an `#![allow(unused)]` to the whole module, because then we can't see which functions or variables are not actually used.

Here I guess this is because of the FIDs as we probably wont use all of them? What I suggest for those situation is to create a new module inside this one and put the FID there, and limiting the `#![allow(unused)]` to that module.

For instance:

```rs

mod sbi {
#![allow(unused)]

/// Keystone EID & FIDs
///
/// See https://github.com/keystone-enclave/keystone/blob/80ffb2f9d4e774965589ee7c67609b0af051dc8b/sdk/include/shared/sm_call.h#L5C1-L6C1
const KEYSTONE_EID: usize = 0x08424b45;

// 1999-2999 are called by host
const CREATE_ENCLAVE_FID: usize = 2001;
const DESTROY_ENCLAVE_FID: usize = 2002;
const RUN_ENCLAVE_FID: usize = 2003;
const RESUME_ENCLAVE_FID: usize = 2005;

// 2999-3999 are called by enclave
const RANDOM_FID: usize = 3001;
const ATTEST_ENCLAVE_FID: usize = 3002;
const GET_SEALING_KEY_FID: usize = 3003;
const STOP_ENCLAVE_FID: usize = 3004;
const EXIT_ENCLAVE_FID: usize = 3006;
}
```

If there is more things which require `#![allow(unused)]` you can add individual `#[allow(unused)]` to those structs or functions. But don't forget to remove them once they become used :)

src/virt.rs
CharlyCst, line 1: This one we will need to discard and then rebase on main (which already include those changes). If you are not sure how to do it let me know, I can help you rebase.

CharlyCst (commented):
Great work on getting all the foundations ready :)

I just suggested a few cleanup, we will also need to rebase on main. Beside that I think it might be a good time to think about setting up the artifact so that we can easily test the code in the CI and on anyone's machine without compiling Keystone locally. I added some pointers in one of my comments, but feel free to ask if you have any question!

config/qemu-keystone.toml
FredKhayat: Okay I see, should I make a pull request on [miralis-artifact-opensbi](https://github.com/CharlyCst/miralis-artifact-opensbi) for that?

src/policy/keystone.rs
FredKhayat: I see, that make sense, I did the changes and added a todo so that I don't forget to remove the `#[allow(unused)]`

config/qemu-keystone.toml
FredKhayat: I removed the whole section `[target.payload]` since it will not be used anymore. 
I started using an OpenSBI+Keystone bundled as one firmware instead of loading `opensbi-jump` and the keystone payload separately.

CharlyCst (approved):
Thanks for making the changes! Looks like a great start foundation :)

I'll rebase and merge

PR #245, opened 2024-10-20
Add linux kernel with protect payload policy.

francois141:
This commit gives the possibility to start Miralis with the linux kernel while the payload is isolated from the firmware through the protect payload policy.

### DIFF
--- a/justfile
+++ b/justfile
@@ -74,13 +74,12 @@ test:
 	cargo run -- run --config {{qemu_virt_hello_world_payload}} --firmware opensbi-jump
 	cargo run -- run --config {{qemu_virt_u_boot_payload}} --firmware opensbi-jump
 
-	# Test benchmark code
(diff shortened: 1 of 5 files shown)


francois141:
@CharlyCst A few values are hardcoded, we need to create an artifact before merging. I will do that

justfile
```diff
@@ -74,14 +74,13 @@ test:
 	cargo run -- run --config {{qemu_virt_hello_world_payload}} --firmware opensbi-jump
 	cargo run -- run --config {{qemu_virt_u_boot_payload}} --firmware opensbi-jump
 
-	# Test benchmark code
+	# Testing with protect payload policy
+	# cargo run -- run --config {{qemu_virt_test_protect_paylod}} --firmware test_protect_payload_firmware
```
francois141, line 78: I will fix that before merging as well

PR #248, opened 2024-10-21
Add linux kernel with protect payload policy.

francois141:
This commit gives the possibility to start Miralis with the linux kernel while the payload is isolated from the firmware through the protect payload policy.

### DIFF
--- a/src/policy/protect_payload.rs
+++ b/src/policy/protect_payload.rs
@@ -2,7 +2,7 @@
 
 use miralis_core::abi_protect_payload;
 
-use crate::arch::pmp::{build_napot, pmpcfg};
+use crate::arch::pmp::pmpcfg;
 use crate::arch::{MCause, Register};
 use crate::host::MiralisContext;
 use crate::policy::{PolicyHookResult, PolicyModule};
@@ -12,8 +12,7 @@ use crate::virt::{RegisterContextGetter, VirtContext};
 pub struct ProtectPayloadPolicy {
     protected: bool,
     general_register: [usize; 32],
-    confidential_values_firmware: ConfidentialCSRs,
-    confidential_values_payload: ConfidentialCSRs,
+    last_cause: MCause,
 }
 
 impl PolicyModule for ProtectPayloadPolicy {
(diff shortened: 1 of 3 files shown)


src/policy/protect_payload.rs
```diff
@@ -103,78 +81,53 @@ impl PolicyModule for ProtectPayloadPolicy {
...
+        mctx.pmp.set_from_policy(1, usize::MAX / 4, pmpcfg::TOR);
+
+        self.last_cause = ctx.trap_info.get_cause();
+
+        /*if !self.protected {
+            return;
+        } else {
+            // todo: In the future, add CSR handling here
+            return;
+        }*/
```
CharlyCst, line 96: Any reason to keep that? If yet let's add a comment to explain why, otherwise let's get rid of it.

src/policy/protect_payload.rs
```diff
@@ -103,78 +81,53 @@ impl PolicyModule for ProtectPayloadPolicy {
...
+            .set_from_policy(1, usize::MAX / 4, pmpcfg::TOR | pmpcfg::RWX);
 
-        // Step 3: Unlock memory
-        mctx.pmp.set_from_policy(0, 0, pmpcfg::INACTIVE);
+        /*if !self.protected {
+            return;
+        } else {
+            // todo: In the future, add CSR handling here
+            return;
+        }*/
```
CharlyCst, line 123: Similar here, should we keep that?

src/policy/protect_payload.rs
```diff
@@ -103,78 +81,53 @@ impl PolicyModule for ProtectPayloadPolicy {
...
             if self.restore_register(i, ctx) {
                 ctx.regs[i] = self.general_register[i];
             }
         }
 
-        // Step 2: Restore sensitives privileged registers
-        self.write_confidential_registers(ctx, true);
+        // Unlock memory
+        mctx.pmp
+            .set_from_policy(0, 0x80400000 / 4, pmpcfg::INACTIVE);
```
CharlyCst, line 114: Ok so this might cause confusion. Why not creating a `build_tor` function? Otherwise someone coming here in the future might wonder why ` / 4`.

CharlyCst (commented):
Just a few nits to clean things up before merging :)

src/policy/protect_payload.rs
francois141: There is a comment inside :=)

src/policy/protect_payload.rs
francois141: We can ignore that, the next commit solve this

src/policy/protect_payload.rs
CharlyCst: Well if you need it for the next commit I suggest you start working on the next commit on this branch and we merge everything once it is working together :) 

Committing intermediate state is not great in general, it might get lost and stay there forever for instance. It's fine to work on a few commits at a time on the same branch and refine them over time.

PR #250, opened 2024-10-21
Improve pmp interface

francois141:
This new interface allow to interact with the pmp module in a more friendly way. It avoids confusion such as forgetting dividing the values by 4 in the pmp entries. Overall it should make the pmp code easier to read and debug.

### DIFF
--- a/src/arch/pmp.rs
+++ b/src/arch/pmp.rs
@@ -7,6 +7,7 @@ use core::fmt;
 use core::fmt::Formatter;
 
 use super::Architecture;
+use crate::arch::pmp::pmpcfg::{INACTIVE, NAPOT, TOR};
 use crate::arch::pmp::pmplayout::{
     ALL_CATCH_OFFSET, DEVICES_OFFSET, INACTIVE_ENTRY_OFFSET, MIRALIS_OFFSET, MIRALIS_TOTAL_PMP,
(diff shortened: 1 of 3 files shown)


src/arch/pmp.rs
```diff
@@ -219,8 +225,28 @@ impl PmpGroup {
...
+        assert!(
+            permissions < 8,
+            "Permissions should not set NAPOT or TOP bits"
+        );
+        self.set(idx, build_tor(until), permissions | TOR);
+    }
+
+    pub fn set_inactive(&mut self, idx: usize, addr: usize) {
+        self.set(idx, build_tor(addr), INACTIVE);
+    }
```
CharlyCst, line 246: Those functions are cool, but they lack some good documentation :)
They will be used in a bunch of places, so having a good doc comment would be great. Can you add an explanation for each of those?

CharlyCst (commented):
Those changes are super cool, thanks for working on those :) 

Could you just add some doc comments for the new functions to set PMP entries? Thanks!

src/arch/pmp.rs
francois141: Done!

francois141:
@CharlyCst Ready to review again!

CharlyCst (approved):
Thanks, I'll rebase and merge :)

PR #251, opened 2024-10-25
Creation of explicit forwarding rules in the protect_payload.rs

francois141:
This commit creates a small framework similar to IPRules that allows to explicitly forwards registers from the payload to the firmware and in the opposite direction. Last but not least, it creates and test a forwarding rule for ecall from S mode.


francois141:
@CharlyCst Rebased commit with your previous comments!

CharlyCst (dismissed):
Thanks for the PR, looks good to me :)
I'll rebase and merge

PR #254, opened 2024-10-26
Run ci / cd pipeline in multicore

francois141:


CharlyCst:
I don't think we should move all of the integration tests to multi-core. Not all of them are designed with multi-core in mind, so they might just be buggy and break down later. For our custom integration test I think single core is better, they are designed to test specific parts of Miralis, and debugging breakage on single core will be much easier.

What we could do is run Linux/OpenSBI on multi-core though, but I would keep all other firmware as is.

francois141:
Got it! I will do the changes

PR #256, opened 2024-10-27
Bugfix: remove wrong bounds check

francois141:
We have a check in the driver code that checks we don't exceed the number of harts determined at compile time. If we run the code on a board with more cores, the code will crash even if the read / write is valid.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -2,8 +2,6 @@
 
 use core::ptr;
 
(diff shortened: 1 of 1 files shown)


CharlyCst:
The hart parameter can come from the firmware (e.g. when writting on the `vclint`), so we have to consider it as untrusted and keep those checks.

PR #257, opened 2024-10-28
Support for full interrupts virtualization

CharlyCst:
This PR implements proper full M-mode interrupts virtualization, as outlined in #186.

Full interrupt virtualization changes the way interrupts are handled: instead of re-injecting physical interrupts into the vCPU (like we use to do until this PR) Miralis maintains the state of the virtual `mip` bits for the virtualized interrupts itself and checks if an interrupt should be injected before each re-entry.

This is an important change because:
1. It is needed for correctness.
2. It will enable out-of-band interrupts, i.e. interrupts managed by Miralis rather than the firmware. This is a requirement for most security policies on multi-core platforms.
3. It should fix the issues we are facing on the board.


CharlyCst:
Current status:

- Implemented full interrupt virtualization for M-mode timer interrupts.
- No support for out-of-band interrupts yet.
- Fixed the SEIE issue on platforms without `Sstc` extension.

To be done:
- Virtualization of MEIE
- Virtualization of MSIE

Both of which requires a virtual CLINT device, which is being developped as part of #199.

PR #258, opened 2024-10-29
Fix Keystone Create Enclave

FredKhayat:
The create enclave function takes as argument a pointer that points to the memory of a lower-privileged mode.
Before accessing that memory, Miralis needs to:

1. Check that the lower-privileged mode has access to that memory.
2. Read from the virtual memory if the lower-privileged mode uses virtual memory.

This PR takes care of that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,6 +461,13 @@ name = "test_helpers"
 version = "0.1.0"
 
 [[package]]
+name = "test_keystone_payload"
+version = "0.1.0"
+dependencies = [
+ "miralis_abi",
+]
+
+[[package]]
 name = "test_protect_payload_firmware"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


CharlyCst:
I rebased this PR following the merge of the previous one, I'll do the review now.

CharlyCst (commented):
Awesome, great work!

I only have a few remarks, one for style and the others to slightly simplify the assembly snippet in the copy bytes. If you want to make the corrections yourself feel free to do so, otherwise I can do it when I come back from the conference. (just note that I rebased your branch, so you might have to reset it to get to the same state as remote).

FredKhayat:
> Awesome, great work!
> 
> I only have a few remarks, one for style and the others to slightly simplify the assembly snippet in the copy bytes. If you want to make the corrections yourself feel free to do so, otherwise I can do it when I come back from the conference. (just note that I rebased your branch, so you might have to reset it to get to the same state as remote).

Thanks for the suggestions! Those are good points, I can make the corrections myself :)

CharlyCst (approved):
Awesome, looks all good to me, thanks for the changes!

PR #259, opened 2024-11-01
Run ci/cd pipeline in a multicore setup

francois141:
Now we run the ci/cd pipeline with a single core. Going multicore make the tests overall more robust to pass as we might encounters race conditions.


francois141:
@CharlyCst This branch is also ready for review. I implemented the changes you ask. It runs the kernel integrations tests in multicore but not the homemade custom firwmares

PR #260, opened 2024-11-01
Implement IPI for the protect payload policy

francois141:
Now when we jump to the payload for the first time, we send an interrupt to all cores and force them to lock the memory even if they are still on the firmware

### DIFF
--- a/src/policy/protect_payload.rs
+++ b/src/policy/protect_payload.rs
@@ -7,6 +7,7 @@ use crate::arch::pmp::pmplayout::POLICY_OFFSET;
 use crate::arch::MCause::EcallFromSMode;
 use crate::arch::{MCause, Register};
 use crate::host::MiralisContext;
+use crate::platform::{Plat, Platform};
 use crate::policy::{PolicyHookResult, PolicyModule};
 use crate::virt::{RegisterContextGetter, VirtContext};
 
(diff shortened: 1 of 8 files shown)


src/policy/protect_payload.rs
```diff
@@ -110,15 +113,26 @@ impl PolicyModule for ProtectPayloadPolicy {
             }
         }
 
-        // TODO: Work on that section
-        // Step 2: Restore sensitives privileged registers
-        /*self.write_confidential_registers(ctx, true);*/
+        if self.first_jump {
+            // Lock memory from all cores
+            broadcast_policy_interrupt();
+        }
```
CharlyCst, line 119: Here you need a proper barrier, because there is no guarantee that the interrupt will be received immediately  (and it practice it wont, there are delays).
If you don't wait for an acknowledgement from all of the cores you might be in a state where the payload is protected only on some cores and thus can be attacked from the remaining unlocked cores.

The simple solution to create a single usage barrier is to have a static atomic counter, you set it once to the number of cores, then each core decrease it by one and wait in a loop for the count to reach 0. You can add a [`core::hint::spin_loop`](https://doc.rust-lang.org/stable/core/hint/fn.spin_loop.html) in the spin loop to calm down the CPU while spinning.

CharlyCst (commented):
So there is one main soundness issue in the PR which is that there is no barrier to wait for all cores to perform the transition when locking the payload. That is a bug that can be exploited in practice (and has been in the past for similar functionalities).

Beside that I have two small nits, some missing doc comments and mixing both `msi` and `ipi` (they are the same things, but MSI is the RISC-V name, and IPI is the Intel name, so let's stick with MSI in this project). I'll fix those small things and merge that PR, but we should create a new one to add the barrier.

PR #263, opened 2024-11-09
Misaligned store loads

francois141:


CharlyCst:
Rebased from #258 since it got merged.

src/policy/protect_payload.rs
```diff
@@ -138,12 +117,129 @@ impl PolicyModule for ProtectPayloadPolicy {
...
+            _ => {
+                panic!("Must be a load instruction here")
+            }
+        }
+
+        ctx.pc += 4;
+        PolicyHookResult::Overwrite
+    }
+
+    fn copy_from_previous_mode(&mut self, src: *const u8, dest: *mut u8) {
```
CharlyCst, line 197: You can't have this function being labelled as safe, it can cause undefined behavior depending on what pointers you pass in. So first thing mark it as unsafe.

Then avoid using pointers like this, instead prefer arrays of bytes `&[u8; 8]` that you can convert to `usize`s using `from_le_bytes` or `to_le_bytes` (see [doc](https://doc.rust-lang.org/std/primitive.u64.html#method.to_le_bytes))

src/policy/protect_payload.rs
```diff
@@ -138,12 +117,129 @@ impl PolicyModule for ProtectPayloadPolicy {
...
+        // Copy the arguments from the S-mode virtual memory to the M-mode physical memory
+        let mode = parse_mpp_return_mode(Arch::read_csr(Csr::Mstatus));
+        unsafe {
+            let slice: *mut [u8] = from_raw_parts_mut(dest, 8);
+            let slice2: &mut [u8] = &mut *slice;
+            Arch::read_bytes_from_mode(src, slice2, mode).unwrap()
+        }
+    }
+
+    fn copy_from_previous_mode_store(&mut self, src: *mut u8, dest: *mut u8) {
```
CharlyCst, line 207: Similar remark here, that should be an unsafe function.

src/virt.rs
```diff
@@ -732,6 +733,13 @@ impl VirtContext {
             MCause::MachineExternalInt => {
                 todo!("Virtualize machine external interrupt")
             }
+            MCause::LoadAddrMisaligned | MCause::StoreAddrMisaligned => {
+                self.emulate_jump_trap_handler()
+            }
+            MCause::InstrAddrMisaligned => {
+                todo!("Implement Instruction Address Misaligned")
+            }
```
CharlyCst, line 741: We shouldn't implement a default behavior in Miralis. If the policies does not handle it, the virtualized firmware should (and that is what happen with OpenSBI when we don't have the protect payload policy).

CharlyCst (commented):
Here are some comments, this is kind of a tricky feature se we will probably need to iterate a bit on it :)

src/arch/metal.rs
```diff
@@ -884,6 +884,65 @@ impl Architecture for MetalArch {
...
+                mprv_filter = in(reg) mstatus::MPRV_FILTER,
+                byte = in(reg) byte_value,
+                success = inout(reg) success,
+                r_mtvec = out(reg) _,
+                )
+            }
+            if success == 0 {
+                Self::write_csr(Csr::Mepc, prev_mepc);
+                Self::write_csr(Csr::Mcause, prev_mcause);
+                Self::write_csr(Csr::Mstatus, prev_mstatus);
```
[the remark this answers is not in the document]
francois141: I guess it is fixed, isn't it?

CharlyCst:
I'll look at that a bit later, but you reverted the rebasing with a force push 🙃

CharlyCst:
So this branch needs to be rebased before it can be reviewed, since the first commit was merged. I can do it after the meeting during my rebase & merge session, but feel free to give it a try if you want to get it done faster :)

PR #264, opened 2024-11-10
Ubuntu, OpenSuse, Fedora, FreeBSD disk images with Miralis

francois141:
This commit provides support to run the ubuntu distribution on top of Miralis by introducing a new payload named ubuntu

### DIFF
--- a/runner/src/artifacts.rs
+++ b/runner/src/artifacts.rs
@@ -13,8 +13,9 @@ use serde::Deserialize;
 
 use crate::config::{Config, Profiles};
 use crate::path::{
-    get_artifact_manifest_path, get_artifacts_path, get_target_config_path, get_target_dir_path,
-    get_workspace_path, is_older,
+    extract_file_extension, extract_file_name, get_artifact_manifest_path, get_artifacts_path,
+    get_target_config_path, get_target_dir_path, get_workspace_path, is_file_present, is_older,
+    remove_file_extention, GZ_COMPRESSION, IMG_EXTENSION, XZ_COMPRESSION, ZST_COMPRESSION,
 };
 use crate::ArtifactArgs;
 
@@ -42,14 +43,55 @@ pub enum Target {
     Payload(String),
 }
 
+// —————————————————————————— Artifact Definitions —————————————————————————— //
+
+trait Artifact {
+    /// Creates an artifact that can be downloaded.
+    fn from_url(name: &str, url: &str) -> Self;
+}
+
+/// A binary artifact, typically a firmware or payload.
 #[derive(Clone, Debug)]
-pub enum Artifact {
+pub enum BinArtifact {
     /// Artifacts that are built from sources.
     Source { name: String },
     /// Artifacts that are downloaded.
     Downloaded { name: String, url: String },
     /// Artifact available as binaries on the local file system.
     Binary { path: PathBuf },
+}
+
+/// A disk artifact
+#[derive(Clone, Debug)]
+pub enum DiskArtifact {
+    /// A disk image that can be downloaded.
+    Downloaded { name: String, url: String },
+}
+
+/// A collection of artifacts
+pub struct AllArtifacts {
+    /// Binary artifacts
+    pub bin: HashMap<String, BinArtifact>,
+    /// Disk artifacts
+    pub disk: HashMap<String, DiskArtifact>,
+}
+
+impl Artifact for BinArtifact {
+    fn from_url(name: &str, url: &str) -> Self {
+        Self::Downloaded {
+            name: name.to_owned(),
+            url: url.to_owned(),
+        }
+    }
+}
+
+impl Artifact for DiskArtifact {
+    fn from_url(name: &str, url: &str) -> Self {
+        Self::Downloaded {
+            name: name.to_owned(),
+            url: url.to_owned(),
+        }
+    }
 }
 
 // ——————————————————————————— Artifact Manifest ———————————————————————————— //
@@ -60,12 +102,21 @@ pub enum Artifact {
 struct ArtifactManifest {
     #[serde(default)]
     bin: HashMap<String, Bin>,
+    disk: HashMap<String, Disk>,
 }
 
 /// Binaries artifacts.
 #[derive(Deserialize, Debug, Default)]
 #[serde(deny_unknown_fields)]
 struct Bin {
+    description: Option<String>,
+    url: Option<String>,
+    repo: Option<String>,
+}
+
+#[derive(Deserialize, Debug, Default)]
+#[serde(deny_unknown_fields)]
+struct Disk {
     description: Option<String>,
     url: Option<String>,
     repo: Option<String>,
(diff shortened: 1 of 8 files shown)


runner/src/run.rs
```diff
@@ -24,6 +24,17 @@ const QEMU_ARGS: &[&str] = &[
...
 
+const UBUNTU_ARGS: &[&str] = &[
+    "-device",
+    "virtio-net-device,netdev=eth0",
+    "-netdev",
+    "user,id=eth0",
+    "-device",
+    "virtio-rng-pci",
+    "-drive",
+    "file=misc/ubuntu-miralis.img,format=raw,if=virtio",
```
CharlyCst, line 35: Where does this file come from? Is there anything special that we need to do to generate it or is it simply a disk image you downloaded somewhere?

runner/src/run.rs
```diff
@@ -46,6 +57,11 @@ pub fn run(args: &RunArgs) -> ExitCode {
         None => return ExitCode::FAILURE,
     };
 
+    // In case we run Ubuntu, we need to download the ubuntu image
+    if args.firmware == "ubuntu" {
+        fetch_ubuntu_image();
+    }
```
CharlyCst, line 63: We don't want to hardcode this, if we need to add more disk or test with different firmware this is not going to scale well.

runner/src/run.rs
```diff
@@ -86,10 +102,10 @@ fn launch_qemu(args: &RunArgs, miralis: PathBuf, firmware: PathBuf) -> ExitCode
         qemu_cmd.arg("-cpu").arg(cpu);
     }
     qemu_cmd
+        .arg("-m")
+        .arg("8G")
```
CharlyCst, line 106: We could extend the config to specify how much memory we want for QEMU. There is already a `[qemu]` table.

runner/src/run.rs
```diff
@@ -180,3 +200,49 @@ fn launch_spike(args: &RunArgs, miralis: PathBuf, firmware: PathBuf) -> ExitCode
...
+        remove_raw_download.arg("ubuntu-24.04.1-preinstalled-server-riscv64.img.xz");
+        remove_raw_download
+            .status()
+            .expect("Failed to remove raw download");
+
+        log::info!("Image ready to use");
+    } else {
+        log::debug!("Ubuntu image already exists.");
+    }
+}
```
CharlyCst, line 248: This is something we want to move in `artifact.rs`, as it deals with downloading stuff.

In its current form that is very specific to Ubuntu, which would create a lot of tech debt as soon as we want to include other disk images. For instance we will probably want to support [Alpine Linux](https://alpinelinux.org/downloads/) for testing (because it is much lighter than Ubuntu).

What I would suggest is to extend the artifact manifest to add a new kind of artifacts (`disk`) and then allow specifying a disk under the `[qemu]` table in the config file. Then we can add some code in the Runner's `artifact.rs` to handle the download from a given URL specified in the artifact manifest.

CharlyCst (commented):
That's supper cool to see Ubuntu work, congrats! :)

I would like to make this future proof, in particular support other disks if we want to in the future and avoid hard-coding any values. (We will most likely use only Ubuntu in the paper eval, but we don't want to accumulate tech debt that will make the project harder to use by other research teams).

I think the best way to go would be to extend the artifact manifest with support for disk artifacts, and then allowing a disk to be specified in the `[qemu]` table of the config. That would make the system much more flexible and future-proof, and also potentially enabling testing in the CI with lightweight distributions.

runner/src/run.rs
francois141: A simple disk image

runner/src/artifacts.rs
```diff
@@ -388,3 +458,71 @@ pub fn list_artifacts(args: &ArtifactArgs) -> ExitCode {
...
+    let file_name: &str = extract_file_name(url_path);
+
+    if !is_file_present(&format!("artifacts/{}-miralis.img", name)) {
+        // Download image
+        log::info!("Disk image not found. Fetching the image...");
+        let mut image_downloader = Command::new("wget");
+        image_downloader.arg(url_path);
+        image_downloader
+            .status()
+            .expect("Failed to downlaod Ubuntu");
```
CharlyCst, line 474: Update error message: "Failed to download disk"

runner/src/artifacts.rs
```diff
@@ -388,3 +458,71 @@ pub fn list_artifacts(args: &ArtifactArgs) -> ExitCode {
...
+
+// ————————————————————————————— Process disk image ————————————————————————————— //
+
+pub fn download_disk_image(name: &str, url_path: &str) -> PathBuf {
+    let file_name: &str = extract_file_name(url_path);
+
+    if !is_file_present(&format!("artifacts/{}-miralis.img", name)) {
+        // Download image
+        log::info!("Disk image not found. Fetching the image...");
+        let mut image_downloader = Command::new("wget");
```
CharlyCst, line 470: I am not sure if `wget` is available everywhere (e.g. on Windows), but good enough for now :)
As a general rule, `curl` is the one that is almost always installed, including on Windows.

runner/src/artifacts.rs
```diff
@@ -388,3 +458,71 @@ pub fn list_artifacts(args: &ArtifactArgs) -> ExitCode {
...
+            todo!("Implement support for extracting files of type .tar.something");
+        }
+
+        // Move and clean
+        log::info!("Moving and cleaning the download");
+        let mut mover = Command::new("mv");
+        mover
+            .arg(filename_without_extension)
+            .arg(format!("artifacts/{}-miralis.img", name));
+        mover.status().expect("Failed to downlaod Disk image");
```
CharlyCst, line 507: There is a Rust function for that in the standard Libary, let's use that instead:
https://doc.rust-lang.org/std/fs/fn.rename.html

runner/src/artifacts.rs
```diff
@@ -388,3 +458,71 @@ pub fn list_artifacts(args: &ArtifactArgs) -> ExitCode {
...
+
+        let mut remove_raw_download = Command::new("rm");
+        remove_raw_download.arg(file_name);
+        remove_raw_download
+            .status()
+            .expect("Failed to remove raw download");
```
CharlyCst, line 513: Same, there is a Rust function for that: https://doc.rust-lang.org/std/fs/fn.remove_file.html

runner/src/artifacts.rs
```diff
@@ -388,3 +458,71 @@ pub fn list_artifacts(args: &ArtifactArgs) -> ExitCode {
...
+    }
+
+    // Now we need to download the u-boot artifact
+    match locate_artifact("u-boot") {
+        Some(Artifact::Downloaded { name, url }) => download_artifact(&name, &url),
+        _ => panic!("U-boot artifact not found"),
+    };
+
+    // We return the u-boot artifact
+    PathBuf::from("artifacts/u-boot")
```
CharlyCst, line 527: Ok so we don't want to hard-code using U-boot here, there are other boot loaders that exits :)

runner/src/main.rs
```diff
@@ -55,6 +55,9 @@ struct RunArgs {
     #[arg(long)]
     /// Path to the configuration file to use
     config: Option<PathBuf>,
+    /// Linux distribution
+    #[arg(long)]
+    distribution: Option<String>,
```
CharlyCst, line 60: Let's call it `disk`, it can contain anything, not only a Linux distribution.

runner/src/path.rs
```diff
@@ -82,3 +82,33 @@ pub fn is_older(a: &Path, b: &Path) -> bool {
         _ => false,
     }
 }
+
+pub const XZ_COMPRESSION: &str = "xz";
+pub const ZST_COMPRESSION: &str = "zst";
```
CharlyCst, line 87: Usually we put constants at the top of the file.

runner/src/artifacts.rs
```diff
@@ -47,6 +48,8 @@ pub enum Artifact {
     Downloaded { name: String, url: String },
     /// Artifact available as binaries on the local file system.
     Binary { path: PathBuf },
+    /// Disk artifact, used to run various linux distributions
+    Disk { name: String, url: String },
```
CharlyCst, line 52: So `Artifact` was intended for `Binary` initially, I don't think the disk abstraction fits well here.

Ideally I would recommend renaming `Artifact` to `BinaryArtifact` and creating a new, separate `DiskArtifact`. That way we can handle them separately, and in particular treat the payload and the disk differently.

CharlyCst (commented):
Super cool, thanks for the changes :)

So one more thing, the disk should be configured not as the payload, but as a new thing called a disk. For that we should add a new item in the `[qemu]` table called disk. The reason is that we might want different payload with different disks, the two are completely independent. (for instance we can boot Linux directly with a disk, without using u-boot). I suggest creating a new `DiskArtifact` and renaming `Artifact` to `BinArtifact` to make it easier to handle both concepts.

We will also need to rebase on top of main, a few things changed in the runner so there might be conflicts.

francois141:
Ready for review!

CharlyCst (commented):
Awesome, the changes look good to me :)

I will merge the PR this evening after the meeting. The next step is to rebase on top of main in order to merge. There might be conflicts, if you want you can give it a try, otherwise I will handle it :)

PR #266, opened 2024-11-18
Implement IPI broadcast for all cores except the emitter.

francois141:
This commit introduce a building block required to go multicore with the ace policy implementation.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -2,6 +2,7 @@
 
 use core::ptr;
 
+use crate::arch::{Arch, Architecture, Csr};
 use crate::config::{self, PLATFORM_NB_HARTS};
 
 pub mod clint {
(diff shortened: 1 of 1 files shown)


src/driver.rs
```diff
@@ -151,4 +152,22 @@ impl ClintDriver {
...
+    /// Create a pending MSI interrupts for each harts of the platform, except the current one.
+    #[allow(dead_code)]
+    pub fn trigger_msi_on_all_other_harts(&mut self) {
+        let current_hart: usize;
+        unsafe {
+            asm!(
+            "csrr {0}, mhartid",
+            out(reg) current_hart
+            );
+        }
```
CharlyCst, line 165: We don't use assembly outside of the `arch` module. This is important as this basically prevent from compiling the module to another architecture, which we need for unit tests and formal verification with the Sail model in the future.

Instead you can use the `read_csr` function from `Arch` to get the current hart ID.

CharlyCst (commented):
See comment :)

src/driver.rs
francois141: I believed it was only for virt.rs. Ma bad, sorry about that! Fixed now!

CharlyCst (approved):

PR #267, opened 2024-11-18
Add support for skipping Spike tests

CharlyCst:
This PR adds support for running Spike tests in the CI as well as skipping tests if the emulator is not available (which avoids breaking tests on user machines without Spike).

A limitations of Spike tests is that they don't work with OpenSBI shutdown as it requires loading OpenSBI as an ELF for Spike to locate the `tohost` symbol, thus there is no Linux Spike test yet. The best option here is most likely to re-compile OpenSBI with a patch to use an ecall to Miralis shuting down on Spike.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,6 +28,12 @@ jobs:
         echo ""
         echo "QEMU version:"
         qemu-system-riscv64 --version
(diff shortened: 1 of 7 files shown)


CharlyCst:
Cool, Spike is finally working on the CI 🙂

PR #271, opened 2024-11-22
Keystone: Run Enclave functionality

FredKhayat:


CharlyCst (commented):
The code looks good to me :)

It seems there is a timeout issue with the CI with one of the keystone test, do you have the same issue locally?

CharlyCst:
Same issue it seems, the tests do not terminate:

```
Boot HART MEDELEG         : 0x000000000000b109
[Info  | miralis::virt] > Hello from test keystone payload
[Info  | miralis::virt] > Enclave created successfully
[Info  | miralis::virt] > Illegal argument test passed
Error: The operation was canceled.
```

The cancelation is the CI timeout. In the test payload are you exiting properly? What is happening when you run the tests in local?

FredKhayat:
> Same issue it seems, the tests do not terminate:
> 
> ```
> Boot HART MEDELEG         : 0x000000000000b109
> [Info  | miralis::virt] > Hello from test keystone payload
> [Info  | miralis::virt] > Enclave created successfully
> [Info  | miralis::virt] > Illegal argument test passed
> Error: The operation was canceled.
> ```
> 
> The cancelation is the CI timeout. In the test payload are you exiting properly? What is happening when you run the tests in local?

When I run the test in local, it terminates successfully, but on the CI, it seems to get stuck when jumping into the enclave. Do you have an idea of what might be the problem?

CharlyCst:
Hmmm in general the QEMU features might be different, which version are you using? The CI is using  6.2.0 if that didn't change recently.
Do you need to pass some special flags?

FredKhayat:
I am using version 7.2.13 and I am not passing any special flags, I will try downgrading my QEMU version to see if I can replicate the bug locally

CharlyCst:
I see, can you check with GDB where the system is stuck? Is it inside of Miralis or somewhere else? And if in Miralis, do you know where?

We didn't have the issue before, so something must have changed ^^

FredKhayat:
I fixed the issue, it happened because I accidentally swapped the names of two CSRs

CharlyCst:
Awesome, good catch ^^

PR #272, opened 2024-11-24
Add formal verification of mret in CI / CD

francois141:
Another test to verify the security monitor!

### DIFF
--- a/src/arch/pmp.rs
+++ b/src/arch/pmp.rs
@@ -228,7 +228,11 @@ impl PmpGroup {
             permissions < 8,
             "Permissions should not set NAPOT or TOP bits"
         );
-        self.set(idx, build_napot(from, to).unwrap(), permissions | NAPOT);
+
+        match build_napot(from, to) {
+            Some(value) => self.set(idx, value, permissions | NAPOT),
+            None => panic!("Invalid arguments for build_napot"),
+        }
     }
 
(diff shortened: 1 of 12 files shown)


francois141:
The first milestone for the automated formal verification is ready! @CharlyCst Can you please make sure the just install pipeline is still easy to use?

CharlyCst:
Cool, it seems like the Ci was failing because Kani is too verbose ^^
Turns out (after some deeper digging) that there is a flag to tune the verbosity, that fixed the issue :)

Before merging this I want to fix a few things, I will come back to it soon.

src/arch/pmp.rs
```diff
@@ -228,7 +228,11 @@ impl PmpGroup {
             permissions < 8,
             "Permissions should not set NAPOT or TOP bits"
         );
-        self.set(idx, build_napot(from, to).unwrap(), permissions | NAPOT);
+
+        match build_napot(from, to) {
+            Some(value) => self.set(idx, value, permissions | NAPOT),
+            None => panic!("Invalid arguments for build_napot"),
+        }
```
CharlyCst, line 235: Out of curiosity why did you need to change that one? It seems semantically equivalent, was Kani complaining about it?

francois141: They are equivalent, I believed Kani complained about that but it was because I was giving an hardcoded illegal value in the kani verify function. You can revert if you want

PR #277, opened 2024-11-25
wip: Fuzzying

francois141:


CharlyCst:
let's not add more and more complexity to the project right away. We should focus one one thing at a time and do it well.

We don't just want to add fuzzing for the sake of fuzzing, and we don't want to add another unrelated toolchain if it doesn't have benefits. let's first focus on getting Kani to work and then evaluate the pros & cons. if Kani is fast enough and cover all we want then fuzzing brings only limited benefits. If Kani has drawbacks or is limited then we might consider fuzzing for extra coverage. We might also want to look into how to limit the overall complexity, for instance with a common abstraction [such as that one](https://crates.io/crates/bolero).

PR #279, opened 2024-11-27
add fid call for the benchmarks

francois141:


francois141:
@CharlyCst The branch is wip, because I need to test it, but you can already review the code :)

PR #281, opened 2024-11-27
SailBugfix: Correct MPP Value in `mstatus` During `mret` Execution

francois141:
This bug fix resolves an issue when bit 20 of the `misa` register is equal to zero, indicating that U-mode is unavailable or disabled. In such cases, the `MPP` field in `mstatus` should be updated to reflect M-mode (`0b11`) instead of incorrectly defaulting to U-mode (`0b00`).


CharlyCst (approved):
Yeay, more bug fixes :)

PR #284, opened 2024-11-28
Add support for model checking with Kani in the CI

CharlyCst:
This PR is based on #272, but integrated model checking in the runner. This make it possible to simply skip model checking for users who did not install Kani. It also uses slightly different names and folder structure to account for potential future extensions.

### DIFF
--- a/model_checking/src/lib.rs
+++ b/model_checking/src/lib.rs
@@ -1,0 +1,42 @@
+use miralis::arch::{mstatus, Arch, Architecture, Mode};
+use miralis::host::MiralisContext;
+use miralis::platform::{Plat, Platform};
+
+#[macro_use]
+mod symbolic;
+mod adapters;
+
+#[cfg_attr(kani, kani::proof)]
+#[cfg_attr(test, test)]
+pub fn mret() {
+    let mpp = match any!(u8) % 3 {
+        0 => 0b00,
+        1 => 0b01,
+        2 => 0b11,
+        _ => unreachable!(),
+    };
+    let mstatus = any!(usize) & !(0b11 << mstatus::MPP_OFFSET);
+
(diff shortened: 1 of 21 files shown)


src/virt/emulator.rs
```diff
@@ -526,9 +531,6 @@ impl VirtContext {
         // The WFI instruction put the processor in a special state that enables taking interrupts
         // even if mstatus.MIE = 0. We keep a bit in the virtual context to model that state.
         self.is_wfi = true;
-        // After a WFI the execution always resumes on the next instruction, whether an interrupt
-        // is taken or not.
-        self.pc += 4;
```
CharlyCst, line 531: Is the moving of the `pc + 4` just refactoring or required for correctness (i.e. did it caused the verification?) @francois141

model_checking/src/lib.rs
```diff
@@ -25,18 +25,46 @@ pub fn mret() {
...
+#[cfg_attr(test, test)]
+pub fn wfi() {
+    let (mut ctx, mut mctx, mut sail_ctx) = get_miralis_symbolic_context();
+
+    ctx.emulate_wfi(&mut mctx);
+
+    execute_WFI(&mut sail_ctx);
+
+    // This field is used only in Miralis. We set it to false otherwise the assertions fails.
+    ctx.is_wfi = false;
```
CharlyCst, line 63: Note: we could also verify that the injection of interrupts is working as expected. There is a `dispatchInterrupt` function in Sail, maybe we could use it somehow for verification as we have a similar function.

CharlyCst (commented):
@francois141 I just have a one question (see bellow), I'll polish that a bit and merge the PR afterward.

PR #287, opened 2024-11-28
Support Sstc extension

FredKhayat:


CharlyCst (approved):
Looks all good, thanks :)

I am not sure how we should handle extensions that can be configured at runtime, Sstc is the first one we implement. Let's start with this and see how that evolves in the future ^^

PR #290, opened 2024-11-28
SailBugfix: mcycle and minstret are not read-only registers

francois141:
Miralis incorrectly treats the mcycle and minstret registers as read-only, which contradicts the official RISC-V specification. This commit resolves the discrepancy to align Miralis's behavior with the specification.


CharlyCst:
I fixed the tests that broke with the change 🙂

PR #298, opened 2024-12-01
SailBugfix: Allow the firmware to write arbitrary values to Mcause

francois141:
This update aligns with the official RISC-V specification, which permits firmware to write arbitrary values to the Mcause register.


francois141:
@CharlyCst Can you please fix the runner for me? I don't want to break anything here :) I know we need to remove one test

CharlyCst:
It seems there is not, Sail does a direct assignment:

```sail
mcause.bits = value
```

And `mcause` is defined to hold any value:

```sail
bitfield Mcause : xlenbits = {
  IsInterrupt : xlen - 1,
  Cause       : xlen - 2 .. 0
}
```

I will merge this PR, but we should keep that in mind in case this surfaces any new bug because of the hardware/model divergence.

PR #302, opened 2024-12-01
SailBugfix: Remove the READ_ONLY_ONE Mask in mideleg and move it to w…

francois141:
…orld_switch.rs

Currently, we enforce the delegation of certain interrupts to S-mode when transitioning to the firmware. However, this behavior is not part of the specification and should therefore be handled in the world_switch.rs code.

### DIFF
--- a/src/virt/csr.rs
+++ b/src/virt/csr.rs
@@ -452,10 +452,7 @@ impl HwRegisterContextSetter<Csr> for VirtContext {
             Csr::Mseccfg => self.csr.mseccfg = value,
             Csr::Mconfigptr => (),                    // Read-only
             Csr::Medeleg => self.csr.medeleg = value, //TODO : some values need to be read-only 0
-            Csr::Mideleg => {
-                self.csr.mideleg = (value & hw.interrupts & !mie::MIDELEG_READ_ONLY_ZERO)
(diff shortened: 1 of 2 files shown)


CharlyCst:
OK so here we also have an issue where there is a divergence between Sail and the hardware we want to expose. For the correctness of interrupts we **have to** enforce the delegation of S-mode interrupts, otherwise we need to virtualize the drivers of devices who can generate S-mode interrupts. This is is compatible with the spec:

> An implementation shall not fix as read-only one any bits of mideleg corresponding to machine-level interrupts (but may do so for lower-level interrupts).

That is what we are exploiting here, we expose a machine with all the supervisor bits delegated.

So long story short, we can't merge this commit without breaking the interrupts. We need to find a way to parametrize the Sail model.

CharlyCst:
It looks to me like the short term solution is to fork the spec and patch it... I will set-up the fork and see if/how we can change the upstream Sail model.

CharlyCst:
I forked the Sail model and pushed a patch there. The model is [here](https://github.com/epfl-dcsl/miralis-sail-riscv/tree/miralis) and the patch is on the `miralis` branch.

I will close this PR for now the tests should pass with the patched model. I will move on and merge the other bugfixes for now.

PR #304, opened 2024-12-01
SailBugfix: mcounteren is read-only except for IR, TMP, and CY fields

francois141:
This commit makes all bits in the mcounteren register read-only, except for the aforementioned IR, TMP, and CY fields, which remain writable.


CharlyCst:
Already merged as part of dbd8d9d4, closing this PR :)

PR #305, opened 2024-12-02
Set sensible interrupt mask in userspace hardware capability

CharlyCst:
This commit sets a interrupt mask with sensible (i.e. that follow the spec) values. Before we had all bits to 1 which caused issues with symbolic execution, namely that the write to MIP was not filtered properly as it should be.


CharlyCst:
@francois141 regarding the `mie` CSR one reason the symbolic execution was failing was the values configured in the `HardwareCapability` struct. I changed the values to something sensible which in turn will filter the proper bits.

There might be other issues, we can look into them once we are done with all other CSRs  🙂

PR #306, opened 2024-12-02
SailBugfix: Make misa read only

francois141:
According to the official risc-v specification, the Misa bit can be read only. This fits the best with the fact that we don't want to explicitly emulate disabling instructions in Miralis.

### DIFF
--- a/firmware/csr_ops/main.rs
+++ b/firmware/csr_ops/main.rs
@@ -306,89 +306,21 @@ fn test_mcause() {
 
 fn test_misa() {
     let misa: u64;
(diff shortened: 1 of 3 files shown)


francois141:
je sais plus si je l'ai déjà envoyé en pr et si tu l'a reject ou pas, désolé si c'est la seconde fois

CharlyCst:
So we could make misa read-only by setting `sys_enable_writable_misa` to false in the `sail-prelude`, is that right?

francois141:
Indeed

PR #308, opened 2024-12-02
SailBugfix: Change only Fiom bit in menvcfg across writes

francois141:
This update ensures that the Fiom bit in the menvcfg register is the only bit we change, aligning with the RISC-V specification.

### DIFF
--- a/src/virt/csr.rs
+++ b/src/virt/csr.rs
@@ -431,6 +431,15 @@ impl HwRegisterContextSetter<Csr> for VirtContext {
                 self.csr.mcounteren = (self.csr.mcounteren & !0b111) | (value & 0b111) as u32
             }
             Csr::Menvcfg => {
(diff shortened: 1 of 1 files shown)


CharlyCst:
Ok so that one is actually important, the bit 63 (`STCE`) controls the supervisor timer extension which we definitely need to enable (otherwise we get a non-negligeable performance penalty due to more traps to the firmware, like we have on the board)

It seems the Sail model is not up to date on that one unfortunately... I suggest we first merge the other PRs and come back to that one at the end, most likely we will need to patch our fork of the sail model to allow the `STCE` bit to be set.

CharlyCst:
@francois141 When you re-generate the spec you might need to add a new bit for that purpose on the Sail spec fork, let me know if you want to look into that together.

PR #309, opened 2024-12-02
SailBugfix: Implement correct behavior of MIE according to the RISC-V…

francois141:
… specification

We verify if there is an interrupt delegation for user mode and adapt the write mask accordingly. The other bits are not overwritten, as specified in the RISC-V documentation.

### DIFF
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -228,6 +228,8 @@ pub mod misa {
     pub const I: usize = 1 << 8;
     /// Integer Multiply/Divide extension
     pub const M: usize = 1 << 12;
+    /// User-Level Interrupts extension
+    pub const N: usize = 1 << 13;
     /// Quad-precision floating-point extension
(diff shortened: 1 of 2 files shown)


francois141:
This is also related to the userspace interrupt feature @CharlyCst, I let you handle this

CharlyCst:
We don't really care about user-level interrupts (it is one of the cases where completeness doesn't matter).

There are two changes which should make the model checking pass without adding support for user-level interrupts:
1. I changed the `HardwareConfig` in the userspace module to set the list of valid interrupt bits (which exlude user-level interrupts). You will get the changes when rebasing on `main`.
2. We can constraint the `misa` to not have the `N` (user-level interrupts, if I remember properly) extension. Now that `misa` is read-only this will disable user-level interrupts for good.

CharlyCst:
I will close for now, let's re-open if updating misa does not fix the issue.

PR #310, opened 2024-12-02
SailFeature: Implement seed register

francois141:
The official Sail RISC-V specification includes the seed register from the cryptography extension, which is currently missing in Miralis. This commit adds support for the seed register.


francois141:
@CharlyCst I am aware it sucks.... Ideally we should have a random value for a non-formal verification workload. Since this feature is so niche, I would say it is enough for the moment

CharlyCst:
The proper way to emulate that is to simply return the value from the hardware using `Arc::read_csr`. But in order to do that we must first detect if the extension is available, and otherwise return a trap. So implementing `seed` properly would be two step:
1. Detect if the extension is available (the spec should say how to do it, we already do that for a few registers)
2. When reading `seed` returns the value from `Arch::read_csr`. Put a constant value in `userspace` but do an actual read in `metal`.

To make the verification work the easy way we could also set `haveZkr` to false in `sail-prelude` or with a custom pass.

src/virt/csr.rs
```diff
@@ -213,6 +213,9 @@ impl RegisterContextGetter<Csr> for VirtContext {
             Csr::Vtype => self.csr.vtype,
             Csr::Vlenb => self.csr.vlenb,
 
+            // Crypto extension
+            Csr::Seed => 0x80000000,
```
CharlyCst, line 217: Which value do you need for the seed in the verification @francois141 ?

I see both `0x80000000` and `0x20000` in two different files?

PR #312, opened 2024-12-02
SailBugfix: Fix write mask for SIP

francois141:
The current Miralis implementation can only modify the SIE field, whereas the RISC-V implementation can also modify the UEIE and USIE fields.


CharlyCst:
Similarly for the user-space interrupt things we can get away by constraining `misa` to not have the N-bit, rather than add support for user-space interrupts everywhere.

CharlyCst:
I will close for now, let's re-open if updating misa does not fix the issue.

PR #316, opened 2024-12-03
SailBugfix: Filter out SIE with MIDELEG bits and provide visibility f…

francois141:
…or userspace interrupt delegation

Our implementation in Miralis overlooks two points. First, we should also check if the MIDELEG bit is set. Second, the official specification allows the software to read the bits for user interrupt delegation.


CharlyCst:
So this is another case of user-level interrupt, I will close the PR as I think it will get solved by disabling user-level interrupts, but we can re-open it if that is not the case.

PR #318, opened 2024-12-03
SailBugfix: Filter out SIP with MIDELEG bits and provide visibility f…

francois141:
…or userspace interrupt delegation

Our implementation in Miralis overlooks two points. First, we should also check if the MIDELEG bit is set. Second, the official specification allows the software to read the bits for user interrupt delegation.


CharlyCst:
This is related to user-level interrupts, this should be solved by disabling the 'N' bit in `misa`.
I will close the PR but we can re-open it if disabling user-level interrupts is not sufficient.

PR #320, opened 2024-12-07
Add logs flags

francois141:


CharlyCst:
Out of curiosity, what do you need this for?
I am not against merging it, but I am just wondering what is the use-case over a standard redirection?

francois141:
It makes the output a bit more clean in my opinion and I use it mainly to filter the SBI calls at the moment :)

PR #323, opened 2024-12-09
Filter out registers per eid/fid in protect payload policy

francois141:
Currently, we allow all a0-a7 filters in the protect payload policy. This commit only allows the arguments register per ecall to pass.


CharlyCst (commented):
Looks cool :)

It would just be nice to have some documentation on top of the python file, just to know what the file is doing and clarify its purpose for people coming-in later.

francois141:
@CharlyCst Ready for review!

PR #333, opened 2024-12-16
Add physical memory protection in Keystone

FredKhayat:


FredKhayat:
For some reason, the keystone test passes locally on qemu 7.2.13 but not on the CI with qemu 6.2.0, I will try to see what is the issue

CharlyCst:
Closing in favor #336, which include this commit plus the modifications that moves the CI to ubuntu 24.04

PR #336, opened 2024-12-16
Keystone PMP + Ubuntu 24.04

FredKhayat:


.github/workflows/ci.yaml
```diff
@@ -3,13 +3,14 @@ name: CI
 on:
   pull_request:
     branches: [ "main", "ci" ]
+  workflow_dispatch:
```
FredKhayat, line 6: I added `workflow_dispatch` so that it becomes possible to run workflows manually, but I can remove it if you prefer. (Before the only way to trigger workflows was to do a pull request on main or ci).

PR #340, opened 2024-12-17
Implement misaligned loads and stores for 2 and 4 bytes.

francois141:
Currently, Miralis only implements support for 8 bytes Misaligned loads and stores. This is not enough to run on the VisionFive2 board. This commit extends the emulation logic.

### DIFF
--- a/firmware/misaligned_op/main.rs
+++ b/firmware/misaligned_op/main.rs
@@ -1,4 +1,3 @@
-//! This firmware acts as a test for misaligned stores and loads. It is only useful in the presence of a platform that doesn't emulate misaligned loads and stores
 #![no_std]
 #![no_main]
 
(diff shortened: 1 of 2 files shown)


firmware/misaligned_op/main.rs
```diff
@@ -8,42 +7,98 @@ use miralis_abi::{log, setup_binary, success};
...
+    ($asm_op:expr, $error_code:expr, $addr:expr, $value:expr) => {
+        unsafe {
+            asm!(
+                $asm_op,
+                addr = in(reg) $addr,
+                r = inout(reg) $value => _,
+            );
+        }
+    };
+}
```
CharlyCst, line 21: What do you gain with this macro? You save `addr = in(reg)` and `r = inout(reg)` only as you need to write the assembly snippet right?

What is the `error_code` about? It doesn't seem used

Overall this macro seems more confusing than it helps.

CharlyCst (commented):
I am not sure what you are trying to achieve with the macro in the test firmware, it is more confusing that it seems to help. Maybe you can either simplify it or just get rid of it?

francois141:
@CharlyCst Can you please review it?

PR #346, opened 2024-12-17
PerformanceFix: Deactivate trace and debug levels at compile time in …

francois141:
…release

Deactivating the logs at compile-time remove many cpu cycles

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -33,6 +33,12 @@ tiny-keccak = { version = "2.0.0", features = ["sha3"] }
 # When running on host architecture as a userspace application, such as when
 # running unit tests.
 userspace = []
+# Define features for Miralis
+release_max_level_trace = ["log/release_max_level_trace"]
+release_max_level_debug = ["log/release_max_level_debug"]
+release_max_level_info = ["log/release_max_level_info"]
+release_max_level_warn = ["log/release_max_level_warn"]
+release_max_level_error = ["log/release_max_level_error"]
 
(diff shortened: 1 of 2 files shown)


src/Cargo.toml
```diff
@@ -21,7 +21,7 @@ path = "main.rs"
 [dependencies]
 uart_16550 = "0.3.0"
 spin = "0.5.2"
-log = { workspace = true }
+log = { workspace = true, features = ["release_max_level_info"] }
```
CharlyCst, line 24: We don't want to disable the logs all the time, otherwise we simply can't log anymore.

What we want to do is to add new features to Miralis that disable the logs based on the config file. The runner has to compute the max log level based on the config, and then pass the appropriate feature to Miralis.

Here is how to disable dependency features from a crate's own features: https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

CharlyCst (commented):
We should control the max log level from the runner, if you are confused about how to do that let me know, I can try to clarify or handle it myself.

CharlyCst:
The PR is good, but by playing a bit with it it turns out that there are some un-anticipated compilation issue: when changing the log level Cargo will not always re-compile due to caching rules with features. That causes some really non-intuitive outcome such as no logs when there should be, or too many logs where they shouldn't.

I implemented another option, compile-time checking, in another PR: #351. I will close this PR in favor of the other, as it achieves a similar result without the re-compilation issues.

PR #350, opened 2024-12-18
Formal verification of interrupt virtualization

francois141:
This commit extends model checking to the virtualization of the interrupts. It builds-up on the previous model checking infrastructure.

### DIFF
--- a/model_checking/src/lib.rs
+++ b/model_checking/src/lib.rs
@@ -1,4 +1,5 @@
-use sail_model::{execute_MRET, execute_WFI};
+use sail_model::{execute_MRET, execute_WFI, trap_handler};
+use sail_prelude::BitVector;
 
 #[macro_use]
 mod symbolic;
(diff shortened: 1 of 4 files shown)


model_checking/src/lib.rs
```diff
@@ -38,3 +41,49 @@ pub fn wfi() {
...
 }
+
+#[cfg_attr(kani, kani::proof)]
+#[cfg_attr(test, test)]
+pub fn interrupt_virtualization() {
+    let (mut ctx, _, mut sail_ctx) = symbolic::new_symbolic_contexts();
+
+    // To avoid an overflow, we artificially set a max limit size for mtvec
+    ctx.csr.mtvec = min(ctx.csr.mtvec, usize::MAX - 10000);
+    sail_ctx.mtvec = BitField::new(min(sail_ctx.mtvec.bits.bits(), (usize::MAX - 10000) as u64));
```
CharlyCst, line 52: Did one of the two implementation get a fault? I am wondering what should be the right behavior, but most likely wrapping add? If the sail implementation does it correctly then we should fix our code :)

francois141: Sail uses a wrapping add, so I do the same now  ^^ Fixed

PR #351, opened 2024-12-18
Enable log filtering at compile time

CharlyCst:
This commit adds helper functions to filter log at compile time. It exposes a `logger::trace_enabled` macro that evaluates to true or false at compile time, thus guaranteed dead code elimination.

We could achieve something similar with the `max_level_*` features from the log crate, however it doesn't work well with incremental compilation: cargo will not re-compile a if a version with a superset of features is already compiled and can be re-used. Thus we opt for compile-time evaluation instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,7 +63,7 @@ fn handle_trap(
     mctx: &mut MiralisContext,
     policy: &mut Policy,
 ) -> ExitResult {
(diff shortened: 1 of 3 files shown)


CharlyCst:
This branch achieves a similar result to #346, but fixes the issue with missing re-compilation.

PR #353, opened 2024-12-19
Formal verification pmp

francois141:


src/virt/mod.rs
```diff
@@ -19,7 +19,7 @@ pub enum ExecutionMode {
 }
 
 /// The context of a virtual firmware.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
```
CharlyCst, line 22: Ok so that's a big NO-GO ^^

Copy is super expensive for big structs, and also super bug prone for anything that is not more than an integer wrapper.

What will happen if we add Copy is that each time we put a `VirtContext` in a new variable the compiler will copy it (instead of using the move semantic). So not only we might have bugs because we update the wrong copy now, but the struct is huge so it will have a huge performance impact in terms of memcopy. 

If the problem is the mutex, see the comment there to fix the issue.

src/virt/mod.rs
```diff
@@ -157,7 +157,7 @@ impl VirtContext {
 }
 
 /// Control and Status Registers (CSR) for a virtual firmware.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
```
CharlyCst, line 160: Same here, Clone is a no-go.

src/arch/userspace.rs
```diff
@@ -31,6 +31,10 @@ static HOST_CTX: Mutex<VirtContext> = Mutex::new(VirtContext::new(
     },
 ));
 
+pub fn return_userspace_ctx() -> VirtContext {
+    *HOST_CTX.lock()
+}
```
CharlyCst, line 36: That smells fishy, because you are not supposed to be able to return a `VirtContext` like that, and indeed after investigation it's because you made VirtContext `Copy`, which we don't want to (see comment there)

There are two options:
1. We make VirtContext `Clone` (but *not* `Copy`), and we clone it there
2. You return a `MutexGuart<VirtContext>` which is basically a wrapper around `VirtContext` but will release the lock when dropped.

I think you probably want option 1 here, so that you release the lock after the clone (otherwise any CSR write with the lock held will deadlock).

model_checking/sail_prelude/Cargo.toml
```diff
@@ -7,3 +7,4 @@ edition = "2021"
 path = "lib.rs"
 
 [dependencies]
+rand = "0.9.0-beta.1"
```
CharlyCst, line 10: We most likely don't want `rand` here.

model_checking/sail_prelude/lib.rs
```diff
@@ -667,7 +684,9 @@ mod tests {
 
         let mut v = BitVector::<SIZE>::new(0);
         let mut val: u64 = 0;
-        for idx in 0..SIZE {
+        for _ in 0..100 {
+            let idx = random::<usize>() % SIZE;
+
```
CharlyCst, line 689: `rand` was added back due to the divergence of the two `sail_prelude`. I can remove it in both versions if you want, just let me know.

CharlyCst (commented):
That's a pretty big feature, but looks like it's going in the right direction :)

I did a first review pass on the technical details, there is one main thing that needs to be fixed around the `Copy` trait in particular.

model_checking/sail_prelude/lib.rs
francois141: Let's keep it like that. One of the tests works with randomness.

model_checking/sail_prelude/Cargo.toml
francois141: In the prelude there is no randomness, it is only used for the unit tests.

model_checking/sail_prelude/lib.rs
CharlyCst: Well I'd prefer not to have randomness, that's why I am asking :)

francois141: But randomness makes the test more robust. Otherwise I have a single pattern. Here I have 100 randoms patterns. Are you sure you don't want to keep it?

francois141: Removed

CharlyCst: Thanks haha, I'm reviewing the PR and I'll merge it, looks great so far :)

CharlyCst:
The verification time is pretty high for that one (700s on my machine), we'll have to see if we can find ways to make it more efficient.

PR #354, opened 2024-12-19
Formal verification of interrupt detection

francois141:


francois141:
@CharlyCst You can already review & merge if you want. I still need to make sure that the detected interrupt is the same. Currently I only check if they detects equivalently but I don't check the returned type

src/virt/emulator.rs
```diff
@@ -782,18 +791,34 @@ fn has_user_mode(ctx: &VirtContext) -> bool {
...
+        None
     }
 }
 
+/// Return the ID of the next interrupt to be delivered, if any.
+fn get_next_interrupt(mie: usize, mip: usize, mideleg: usize) -> Option<usize> {
+    // First function
+    let ip = mie & mip & !mideleg;
+
+    // Second function (they should be equivalent)
```
CharlyCst, line 818: What does that mean 'thay should be equivalent'?

src/virt/emulator.rs
```diff
@@ -843,8 +868,7 @@ mod tests {
         assert_eq!(get_next_interrupt(0b000, 0b010, 0b000), None);
         assert_eq!(get_next_interrupt(0b010, 0b010, 0b010), None);
 
-        assert_eq!(get_next_interrupt(0b001, 0b001, 0b000), Some(0));
```
CharlyCst, line 846: This one we can keep, but yes the next one we have to update it with the new order.

model_checking/src/symbolic.rs
```diff
@@ -121,8 +121,9 @@ pub fn new_ctx() -> VirtContext {
     // We fix the architecture type to 64 bits
     ctx.csr.misa = (0b10 << 62) | (ctx.csr.misa & ((1 << 62) - 1));
 
-    // We must have support for usermode in Miralis
+    // We must have support for usermode and in Miralis
```
CharlyCst, line 124: and ? Typo here^^

Is there an issue if there is no S mode? Miralis should support both with and without S mode (but he hardware capabilities must be updated accordingly I think?)

CharlyCst (commented):
Here are a few comments.

In addition just to get a better idea, how hard would it be to verify the two functions combined? It seems like you are almost there, but the Sail spec might not have an equivalent function? I am not sure.

src/virt/emulator.rs
francois141: It was for me to debug the formal verification (aka François' brain dump)

model_checking/src/symbolic.rs
francois141: Currently the interrupt virtualisation process won't work if we don't have support for S-mode. Let me handle this in a second PR. The core of the logic is there and can be merged.

francois141:
@CharlyCst I rebased the commit and introduced a new function in Virt-sail which allows us to model check the entire virtualisation of the interrupts at once :)

PR #355, opened 2024-12-22
Formal verification of the entire interrupt virtualization pipeline

francois141:
This commit extends model checking to the entire interrupt virtualization pipeline. It builds-up on the previous model checking infrastructure.

The famous branch that got force-pushed :)

### DIFF
--- a/model_checking/sail_model/lib.rs
+++ b/model_checking/sail_model/lib.rs
@@ -393,6 +393,10 @@ pub fn hex_bits_forwards_matches<const N: usize>(
     true
 }
 
+pub fn get_config_print_instr(sail_ctx: &mut SailVirtCtx, unit_arg: ()) -> bool {
+    false
+}
+
 pub fn get_config_print_reg(sail_ctx: &mut SailVirtCtx, unit_arg: ()) -> bool {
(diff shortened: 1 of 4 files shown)


CharlyCst:
@francois141 I rebased the PR as it was targeting an old branch. You can double check that everything looks OK.

Just one question before merging this, where does the `step_interrupt_only` function? It seems you created it from `step`, so you added it to the custom Sail model?

Otherwise looks pretty cool, so now we can check interrupts end-to-end :)

francois141:
Indeed. The step function is something like

```C
if interrupt_ready() {
fire
} else {
fetch_decode_execute()
}
```

I replace with 

```C
if interrupt_ready() {
fire
} else {
 // Nothing 
}
```

CharlyCst:
Where did you do the change? On the Sail model or directly on the Rust code? How can I reproduce it?

If it is on the Sail model, can you add it on the repo (https://github.com/epfl-dcsl/miralis-sail-riscv/tree/miralis) ?

francois141:
@CharlyCst I pushed the modifications on the sail mode, done

PR #360, opened 2024-12-24
Write csr verification

francois141:


francois141:
@CharlyCst I solved at the same time a csrRead (which was only adding the LCOFIE filter in one of the csr register)

PR #362, opened 2024-12-26
Add offset pmp verification

francois141:
@CharlyCst Two points

First of all, I am aware I don't check all offsets. Even after 30 minutes the verification was not ending. Therefore I only formally verify the offset of the default policy (which is ultimately the one we want). 

Second, I fixed a bug in userspace.rs and we have the exact same one in metal.rs. I plan to fix it in a next commit. Also I see that we have some logic in userpsace.rs. I will refactor it to squash it with the metal.rs logic. The motivation is to increase the coverage of the symbolic verification.

### DIFF
--- a/model_checking/src/lib.rs
+++ b/model_checking/src/lib.rs
@@ -7,7 +7,7 @@ use sail_model::{
     execute_MRET, execute_WFI, pmpCheck, readCSR, step_interrupts_only, writeCSR, AccessType,
     ExceptionType, Privilege,
 };
-use sail_prelude::{BitField, BitVector};
+use sail_prelude::{sys_pmp_count, BitField, BitVector};
 
 use crate::adapters::{pmpaddr_sail_to_miralis, pmpcfg_sail_to_miralis, sail_to_miralis};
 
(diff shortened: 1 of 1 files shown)


CharlyCst:
Sounds good for the offset. Overall the PMP verification takes forever because we verify the 64 PMP entries, if we could verify only 16 it would be much faster. The problem is that Sail checks in the number of PMPs in the decoder, not in the read/write CSR function. I'm not sure what is the cleanest way to reduce the coverage to platforms with 16 PMPs to make the verification faster.

For the user-space bug, why + 7? What was the issue?

francois141:
@CharlyCst Let's merge this branch after the deduplicating one. I will rebase it on top of it

francois141:
I added the +7 in case the number of pmp is not a divisor of 8, but I realise this is not a feasible scenario. Let me remove it. Let's merge the feature (it is still faster than the previous commit) and we can check as a next step for 16 pmp. To be honest, I don't think it is a big deal

francois141:
@CharlyCst Rebased and ready for review!

CharlyCst (approved):

PR #363, opened 2024-12-26
Deduplicate set_mpp and get_raw_faulting_instr

francois141:
Both instructions performs exactly the same logic in userspace.rs and metal.rs and are not architecture dependent. This commit removes the smell

### DIFF
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -71,7 +71,6 @@ pub trait Architecture {
     /// environment.
     unsafe fn set_csr_bits(csr: Csr, bits_mask: usize) -> usize;
 
-    unsafe fn write_pmp(pmp: &PmpGroup) -> PmpFlush;
     unsafe fn sfencevma(vaddr: Option<usize>, asid: Option<usize>);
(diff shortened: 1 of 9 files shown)


src/arch/mod.rs
```diff
@@ -517,3 +510,34 @@ impl From<usize> for Width {
...
+        }
+    }
+
+    let instr_ptr = trap_info.mepc as *const u32;
+
+    // With compressed instruction extention ("C") instructions can be misaligned.
+    // TODO: add support for 16 bits instructions
+    let instr = ptr::read_unaligned(instr_ptr);
+    instr as usize
+}
```
CharlyCst, line 544: This will fault when run in user-space. We probably won't run into those kinds of issues until we verify the instruction fetch, so its fine for now, but lets keep that in mind if it shows up in the future.

CharlyCst (approved):
Cool, thanks for the cleanup!

src/arch/mod.rs
francois141: @CharlyCst Good to know. This commit was a simple refactoring and doesn't change the userspace function. Let me check if I find a way to formally verify the decoder

PR #364, opened 2024-12-27
Optimize log handling on the world switch critical path by eliminatin…

francois141:
…g dead code.

Currently, Miralis checks whether each log should be displayed on every world switch, which results in unnecessary cycle loss. This commit introduces a macro to eliminate unused log checks, optimizing performance in the critical path.

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -148,7 +148,44 @@ macro_rules! trace_enabled {
     };
 }
 
-pub(crate) use trace_enabled;
+/// Returns true if the Debug logging level is enabled for the current module.
+///
+/// This macro is guaranteed to be executed at compile time, thus dead-code elimination is
+/// guaranteed by any reasonable compiler.
+macro_rules! debug_enabled {
+    () => {
+        // We wrap the computation in a const block to guarantee execution at compile time
+        const { crate::logger::enabled(core::module_path!(), log::Level::Debug) }
+    };
+}
+
+/// Log a message at the trace level.
+///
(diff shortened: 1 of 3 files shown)


src/logger.rs
```diff
@@ -148,7 +148,36 @@ macro_rules! trace_enabled {
...
+/// guaranteed by any reasonable compiler.
+macro_rules! debug_enabled {
+    () => {
+        // We wrap the computation in a const block to guarantee execution at compile time
+        const { crate::logger::enabled(core::module_path!(), log::Level::Debug) }
+    };
+}
+
+/// Wrappers around the normal logs that allow to dead-code optimize automatically the logs
+/// The motivation beyond this is to cut the latency in the trap. Currently, the logs are responsible for most of the latency
```
CharlyCst, line 163: As a general remark: this is a doc comment, the goal is to explain how and when to use the macro. It is the text that will show up in your IDE when you hover on top of the macro, therefore you should add the doc comment to both `trace` and `debug`.
A good rule of thumb for writing doc comment is to start with a single one-line to tell what the function/macro does, then skip a line and explain in more details if needed. And don't hesitate to copy-paste doc comment on each function.

Those are different from "standard" comments with `//` which is where you would explain general stuff about the code that you don't expect to show up in the doc. Those are more appropriate to explain motivation for instance.

CharlyCst (approved):
Looks good to me, I'll just patch the comments a bit and merge it :)

PR #365, opened 2024-12-27
Parametrize Benchmark Type at Compile Time

francois141:
Currently, we only support a single type of benchmark. However, we are planning to introduce lightweight benchmarks with a smaller CPU footprint. This commit adds functionality to parametrize the benchmark type, similar to how architecture, platform, and policies are handled. This change lays the foundation for future extensions to support additional benchmark types.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -66,8 +66,8 @@ pub const DELEGATE_PERF_COUNTER: bool = is_enabled_default_false!("MIRALIS_DELEG
 pub const PLATFORM_BOOT_HART_ID: usize =
     parse_usize_or(option_env!("MIRALIS_PLATFORM_BOOT_HART_ID"), 0);
 
-/// Whether any benchmark is enabled
-pub const BENCHMARK: bool = is_enabled_default_false!("MIRALIS_BENCHMARK");
+/// What kind of benchmark we want to have in Miralis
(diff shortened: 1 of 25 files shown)


src/config.rs
```diff
@@ -67,7 +67,7 @@ pub const PLATFORM_BOOT_HART_ID: usize =
     parse_usize_or(option_env!("MIRALIS_PLATFORM_BOOT_HART_ID"), 0);
 
 /// Whether any benchmark is enabled
```
CharlyCst, line 69: Small nit: the doc comment should be updated.

CharlyCst (commented):
So you want to enable selecting a module for the benchmark infrastructure, right?

Overall sounds good to me. Just to have a better idea of where you are going, what do you want to change in another benchmark module? The kinds of events you count, how you count them, or something else?

src/config.rs
francois141: :)

francois141:
You commnented the last WIP pr. The goal of this pr is to introduce this new benchmark. The problem with this one is that it tracks the number of cycles spent in Miralis which has some overhead in the cycles. We don't want this overhead for some of the benchmarks

francois141:
@CharlyCst I fixed the last comment and rebased. Feel free to review again

CharlyCst (approved):

PR #366, opened 2024-12-27
Add counter benchmark

francois141:


src/benchmark/counter.rs
```diff
@@ -0,0 +1,52 @@
+use core::sync::atomic::{AtomicUsize, Ordering};
+
+use crate::benchmark::default::IntervalCounter;
+use crate::benchmark::Counter::{FirmwareExits, WorldSwitches};
+use crate::benchmark::{BenchmarkModule, Counter, Scope};
+
+pub struct CounterBenchmark {}
+
+static NB_WORLD_SWITCHES: AtomicUsize = AtomicUsize::new(0);
+static NB_FIRMWARE_EXITS: AtomicUsize = AtomicUsize::new(0);
```
CharlyCst, line 10: This won't fly in a multi-core setup, all the cores would increment the same counters. We probably want to track the counters on a per-core bassis (we might run some benchmarks on a subset of cores for instance)

francois141: Oh, I thought the goal was to get the total number of world switches overall. Looks good to me then. I will apply the changes

PR #367, opened 2024-12-31
The first step in the formal verification of the decoder.

francois141:
This commit establishes the foundation for the formal definition of the decoder.


francois141:
@CharlyCst This commit should be merged after the decoder optimisation

PR #372, opened 2025-01-04
Measure the Cost of Traps, World Switches, and Misaligned Operat…

francois141:
…ions with the Protect Payload Policy

This commit extends the tracing firmware with three new latency measurements in the Protect Payload policy and ensures that no additional performance losses are introduced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -551,6 +551,7 @@ dependencies = [
 name = "tracing_firmware"
 version = "0.1.0"
 dependencies = [
+ "config_helpers",
  "miralis_abi",
 ]
 
--- a/config/test/spike-latency-benchmark-protect-payload.toml
+++ b/config/test/spike-latency-benchmark-protect-payload.toml
@@ -1,0 +1,23 @@
+# A simple configuration to run on the Spike platform
+
+[log]
+level = "info"
+color = true
+
+[vcpu]
+max_pmp = 8
+delegate_perf_counters=true
+
+[platform]
+name = "spike"
+
+[target.miralis]
+# Build profile for Miralis (dev profile is set by default)
+profile = "release"
+
+[target.firmware]
+# Build profile for the firmware (dev profile is set by default)
+profile = "release"
+
+[policy]
+name = "protect_payload"
--- a/firmware/tracing_firmware/Cargo.toml
+++ b/firmware/tracing_firmware/Cargo.toml
@@ -11,3 +11,4 @@ path = "main.rs"
 
 [dependencies]
 miralis_abi = { path = "../../crates/abi" }
+config_helpers = { path = "../../crates/config_helpers" }
--- a/firmware/tracing_firmware/main.rs
+++ b/firmware/tracing_firmware/main.rs
@@ -9,9 +9,14 @@
 
 use core::arch::{asm, global_asm};
 
+use config_helpers::parse_str_or;
 use miralis_abi::{failure, log, setup_binary, success};
 
 setup_binary!(main);
+
+const POLICY_NAME: &str = parse_str_or(option_env!("MIRALIS_POLICY_NAME"), "default_policy");
+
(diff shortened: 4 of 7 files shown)


francois141:
@CharlyCst Let's first merge this pr and then the other. So we can see the cycle gain for the decoder

crates/tracing/src/lib.rs
CharlyCst, line 1: I am writing some comments here, but as explained in the main review feedback we can simply get rid of this library by having a single tracing firmware and keeping things simple.

So the comments on this file are simply informational.

crates/tracing/src/lib.rs
```diff
@@ -0,0 +1,220 @@
+//! Miralis ABI
+//!
+//! While miralis forwards standard SBI calls to the virtualized firmware, miralis do expose its own
+//! ABI for the firmware to interact with.
+//!
+//! The Miralis ABI tries to be compatible with the SBI specification as best as it can.
+//! See: https://github.com/riscv-non-isa/riscv-sbi-doc
```
CharlyCst, line 7: This is a copy paste from another library, it doesn't explain what this library is doing nor how it should be used. The current doc comment might cause confusion and make the crate less maintainable.

crates/tracing/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+#[macro_export]
+macro_rules! measure {
+    ($policy_name:literal, $out:literal, $in:expr, $in2:expr) => {
+        let mut values: [usize; NB_REPEATS] = [0; NB_REPEATS];
+
+        for i in 0..NB_REPEATS {
+            values[i] = $in();
+        }
+
+        let stats = get_statistics(values);
```
CharlyCst, line 27: This macro uses locally-defined function, so it will fail with a compile error when used by another crate or module. There is a simple way to fix this, which is to use an absolute path for each local function used. There is a magic variable `$crate` to do that:

```rs
let stats = $crate::get_statistics(values)
```

With the above version the macro can be used without importing the `get_statistics` function. In general macros in libraries should encapsulate all of their import so that they work like standard functions.

crates/tracing/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+.global _empty_handler
+_empty_handler:
+    // Skip illegal instruction (pc += 4)
+    csrrw x5, mepc, x5
+    addi x5, x5, 4
+    csrrw x5, mepc, x5
+    // Return back to OS
+    mret
+"#,
+);
```
CharlyCst, line 61: In libraries we don't want to insert global assembly, because it can override existing symbols and cause at best linker errors, at worst arbitrarily bad things.

CharlyCst (changes requested):
Thanks for the changes, it's cool to have more stats :)

So turns out `tracing_firmware` and `tracing_firmware_protect_payload` are basically the same file at the exception of the name of the policy according to `diff`:

```diff
➜  firmware git:(851f56c5) ✗ diff tracing_firmware/main.rs tracing_firmware_protect_payload/main.rs
36c36
<         "Default",
---
>         "Protect",
74c74
<         "Default",
---
>         "Protect",
80c80
<         "Default",
---
>         "Protect",
```

But can simply read the policy name from the config, as we do in Miralis:

```rs
pub const POLICY_NAME: &str = parse_str_or(option_env!("MIRALIS_POLICY_NAME"), "default_policy");
```

(if the firmware don't have access to it then we should fix that and expose the environment variables too)

Using this we can simply have a single tracing firmware (and so we don't need the library anymore), which will make things much easier to maintain in the long run :)

francois141:
@CharlyCst Thank you for all the nice and relevants comments. I did the changes. Feel free to review again ;)

CharlyCst (approved):
Thanks for the changes, looks good to me :)

PR #375, opened 2025-01-09
Remove Ecall and Ebreak from Instr

francois141:
The ECALL and EBREAK instructions have distinct mcause values and are not considered illegal instructions. Since they trigger different types of traps, they do not require decoding in our decoder for illegal instructions. As a result, we can safely remove them from the instruction decoder.


CharlyCst (commented):
I rebased the PR. Unfortunately it breaks the `verify_decoder` Kani harness because of the removed `Ecall` and `Ebreak` instructions. Did you already fix this issue in another PR or branch?

francois141:
Yes, I fixed this in another branch. I rebased the code with a decoder that should work now.

PR #378, opened 2025-01-10
Split instr struct

francois141:


francois141:
@CharlyCst Can we please merge this on Monday? This optimisations allows us to gain a few cycles in the misaligned trap and stores. I suspect it will have a small influence on the final results. Therefore we need to merge it asap before the measurements.

CharlyCst:
The PR has conflicts, so it can't be merged as is. Also I would like to check numbers before merging this. Splitting the instructions is a pretty big changes which might bite us back down the line, se we need to know what tradeoff we are taking here.

francois141:
It makes us win around 20 cycles in the case of a firmware trap and 70 for a world switch. Splitting load/ stores with the illegal instructions is a logic step as those never come together and arrive in disjoint scenario. Loads and store is similar except we increase a bit the size of MIralis because we need to unroll this. In any cases, let's have a talk about that asap such that I know if I should measure Miralis with this or not ^^

francois141:
Yes I can rebase. The intuition why we win some cycles is because you don't have to dispatch between two types and you remove a few match statements

francois141:
@CharlyCst I will create a new PR as rebasing this one it too complicated. It would be nice in the future to review the PR and merge them in linear order to avoid this kind of problems

PR #379, opened 2025-01-15
SailBugfix: Illegal instruction decoder decodes too many instructions

francois141:
Currently, the illegal instruction decoder decodes too many instructions instead of returning an unknown instructions. This commit fixes the issues and produces a formally verified decoder.

### DIFF
--- a/model_checking/src/lib.rs
+++ b/model_checking/src/lib.rs
@@ -525,7 +525,7 @@ pub fn verify_decoder() {
     let (_, mctx, mut sail_ctx) = symbolic::new_symbolic_contexts();
 
     // Generate an instruction to decode
-    let instr = any!(u32);
+    let instr = any!(u32, 0x30001073);
 
     // Decode values
     let decoded_value_sail = ast_to_miralis_instr(encdec_backwards(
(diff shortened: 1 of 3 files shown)


CharlyCst:
This is a duplicate of #375, closing the PR.

francois141:
@CharlyCst nonono ^^ This is a sail bug fix :)

francois141:
```rust
                _ if func7 == 0b0010001 => {
                _ if func7 == 0b0010001 && (raw & 0b111111111111111) == 0b000000001110011 => {
``

This is part of a bugfix for example

CharlyCst:
What do you need to match then? the `(raw & 0b111111111111111) == 0b000000001110011` looks like a match on the immediate, is that the case or are the other bits you need to compare?

francois141:
@CharlyCst Some bits needs to be explicitly equal 0. At least the sail model says this

CharlyCst:
It looks to me like this bit was cleaner (and more efficient) before removing the `EBREAK` and `ECALL`, so I am trying to understand what is going on here. We already parse all the sections of the raw instructions, which section needs 0s then? Can we just check conditions on those parts rather than re-parse the `raw`? 
We are always checking for the same value (`0b000000001110011`) so it seems we could use a nested match or something like this too.

francois141:
It was cleaner but also wrong! rd must be equal 0 for some instructions as those bits are used nowhere for the instruction.

francois141:
If you want I can write in a cleaner way

francois141:
@CharlyCst I rewrote it in a readable way

CharlyCst:
Thanks for the changes, looks good now!

I rebased and fixed a unit test :)

PR #383, opened 2025-01-15
SailBugfix: Modifies only mie in write Sie when the interrupt is dele…

francois141:
…gated

Before modifying each specific interrupt enable bit (e.g., SEIE, STIE, or SSIE), the code checks whether the corresponding bit in Mideleg is set. Mideleg determines whether a specific interrupt is delegated to the supervisor level, so this ensures that only interrupts managed by the supervisor are updated.

### DIFF
--- a/src/virt/csr.rs
+++ b/src/virt/csr.rs
@@ -541,10 +541,9 @@ impl HwRegisterContextSetter<Csr> for VirtContext {
                 );
             }
             Csr::Sie => {
-                // Clear S bits
-                let mie = self.get(Csr::Mie) & !mie::SIE_FILTER;
-                // Set S bits to new value
-                self.set_csr(Csr::Mie, mie | (value & mie::SIE_FILTER), mctx);
+                // Only delegated interrupts can be enabled through `sie`
+                let mideleg = self.get(Csr::Mideleg);
+                self.csr.mie = (self.csr.mie & !mideleg) | (mideleg & value);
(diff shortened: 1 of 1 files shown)


CharlyCst:
I see what you want to do here, shouldn't it simply be:

```rs
// Only delegated interrupts can be enabled through `sie`
let mideleg = self.get(Csr::Mideleg);
self.csr.mie = (self.csr.mie & !mideleg) | (mideleg & value);
```

where all delegated interrupts can be modified?

francois141:
The two formulations are not equivalent. In the case mideleg is set to 0 you don't want to change the value of the bit. Therefore we need three if statements ^^

francois141:
The problem is that if only one of the interrupt is delegated, with the previous implementation we can clear the entire mask which is wrong. Regarding lcofie, let me fix that

francois141:
@CharlyCst Ready for review! I did the change

CharlyCst:
Yes, I agree the previous implementation was wrong, but is:

```rs
// Only delegated interrupts can be enabled through `sie`
let mideleg = self.get(Csr::Mideleg);
self.csr.mie = (self.csr.mie & !mideleg) | (mideleg & value);
```

Correct? That was my question ^^

francois141:
This also looks correct, let me test with kani

francois141:
Works, pushed

PR #386, opened 2025-01-16
Cleanup Keystone code

FredKhayat:
- Cleanup Keystone policy
- Add keystone artifacts
- Give more memory to Qemu (so that the Keystone tests can run)


CharlyCst (approved):
Looks good to me, thanks for the changes!

I'll rebase and merge

PR #387, opened 2025-01-21
Board premier 550

francois141:


src/driver/uart.rs
```diff
@@ -0,0 +1,50 @@
+//! UART Driver
+//!
+//! This module implements a driver for the UART driver. This driver allows Miralis to communicate with the UART.
```
CharlyCst, line 3: Typo: `a driver for the UART driver`

src/driver/uart.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn is_line_busy(&mut self) -> bool {
+        // Line Status Register
+        const LSR_OFFSET: usize = 0x05;
+        // Transmit Holding Register Empty
+        const LSR_THRE: u8 = 0x20;
+
+        unsafe { ptr::read_volatile(self.get_register(LSR_OFFSET) as *const u8) & LSR_THRE == 0 }
+    }
+}
+impl Write for UartDriver {
```
CharlyCst, line 43: We add a blank line between two implementations block.

```rs
}

impl Write for UartDriver {
```

src/driver/uart.rs
```diff
@@ -0,0 +1,50 @@
...
+impl UartDriver {
+    pub const fn new(serial_port_base_addr: usize, size_per_register: usize) -> Self {
+        UartDriver {
+            serial_port_base_addr,
+            size_per_register,
+        }
+    }
+
+    pub const fn get_register(&mut self, offset: usize) -> usize {
+        // Registers are 32 bits wide on the board
```
CharlyCst, line 22: What does this comment mean? Aren't the register size read of `self.size_per_register`?

src/main.rs
```diff
@@ -35,6 +35,7 @@ extern "C" {
 pub(crate) extern "C" fn main(_hart_id: usize, device_tree_blob_addr: usize) -> ! {
     // On the VisionFive2 board there is an issue with a hart_id
     // Identification, so we have to reassign it for now
+
```
CharlyCst, line 38: Let's not add newlines in this un-related file.

src/main.rs
```diff
@@ -52,6 +53,7 @@ pub(crate) extern "C" fn main(_hart_id: usize, device_tree_blob_addr: usize) ->
     log::debug!("DTS address: 0x{:x}", device_tree_blob_addr);
 
     log::info!("Preparing jump into firmware");
+
```
CharlyCst, line 56: Same, no newline here.

src/platform/mod.rs
```diff
@@ -22,9 +23,10 @@ use crate::{device, logger};
 pub type Plat = select_env!["MIRALIS_PLATFORM_NAME":
     "miralis"     => miralis::MiralisPlatform
     "visionfive2" => visionfive2::VisionFive2Platform
+    "premierp550"             => premierp550::PremierP550Platform
```
CharlyCst, line 26: Minor nit: align the '=>' properly. Is it a tab on your machine displayed with a different width on your machine?

src/platform/mod.rs
```diff
@@ -22,9 +23,10 @@ use crate::{device, logger};
...
     _             => virt::VirtPlatform
-];
 
+];
```
CharlyCst, line 29: The newline moved, we want the newline after the `];`

src/platform/premierp550.rs
```diff
@@ -0,0 +1,234 @@
...
+use crate::driver::uart::UartDriver;
+use crate::Platform;
+
+// —————————————————————————— Platform Parameters ——————————————————————————— //
+
+// TODO: Check about that if it is the case or not
+const MIRALIS_START_ADDR: usize = TARGET_START_ADDRESS;
+const FIRMWARE_START_ADDR: usize = TARGET_FIRMWARE_ADDRESS;
+
+// TODO: What is the base of the clint device?
```
CharlyCst, line 24: Just a note: this can be found in the SoC manual (so the Eswin [EIC7700X](https://www.sifive.com/document-file/eic7700x-datasheet))

src/platform/premierp550.rs
```diff
@@ -0,0 +1,234 @@
...
+        size: CLINT_SIZE,
+        name: "CLINT",
+        device_interface: &VIRT_CLINT,
+    },
+    VirtDevice {
+        start_addr: TEST_DEVICE_BASE,
+        size: TEST_DEVICE_SIZE,
+        name: "TEST",
+        device_interface: &VIRT_TEST_DEVICE,
+    },
```
CharlyCst, line 60: We can remove the test device on the board, it consumes one PMP and we don't have that many.

src/platform/premierp550.rs
```diff
@@ -0,0 +1,234 @@
...
+
+// ———————————————————————————————— Platform ———————————————————————————————— //
+
+pub struct PremierP550Platform {}
+
+impl Platform for PremierP550Platform {
+    // TODO: Change this number
+    const NB_HARTS: usize = 5;
+    // TODO: Check this number
+    const NB_VIRT_DEVICES: usize = 2;
```
CharlyCst, line 71: You can do `VIRT_DEVICES.len()`, or maybe `const { VIRT_DEVICES.len() }`.

src/platform/premierp550.rs
```diff
@@ -0,0 +1,234 @@
...
+    },
+];
+
+// ———————————————————————————————— Platform ———————————————————————————————— //
+
+pub struct PremierP550Platform {}
+
+impl Platform for PremierP550Platform {
+    // TODO: Change this number
+    const NB_HARTS: usize = 5;
```
CharlyCst, line 69: So the core numbering starts at 0 or 1? On the vision five we do have 5 cores but one of them is smaller, I am not sure if it is the case here?

We can figure that out later, but let's just keep it in mind :)

config/premierp550.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[debug]
+# max_firmware_exits = 2000
+
+[vcpu]
+max_pmp = 0
+
+[platform]
+name = "premierp550"
+nb_harts = 5
```
CharlyCst, line 15: Do we know if we have 4 or 5 cores? We can change that later, but it would be interesting to know. The VisionFive has 5, but one is smaller than the main 4, I am not sure if the P550 is similar.

src/driver/uart.rs
CharlyCst: Ho I see, that code comes from the VisionFive2, we can simply remove the comment then :)

CharlyCst (changes requested):
Thanks for the patch, great changes, nice code and comments/commit messaged!

I mostly have stylistic remarks and clarifications, once fixed we are good to merge. If you need help applying patches to the right commits let me know.

CharlyCst:
Merged as part of #409, thanks for P550 platform support!

PR #389, opened 2025-01-23
Update linux versions in artifacts.toml

francois141:
The new versions contains the benchmark module that can be used to retrieve some statistics from Miralis in the linux kernel.


CharlyCst:
The test broke, can you try to see if that works locally?

CharlyCst:
Closing for inactivity

PR #390, opened 2025-01-25
Formally verify the trap virtualization process

francois141:
Currently, we only verify formally the virtualization process of interrupt. This commit extends the logic to the traps.

@CharlyCst I am completely aware you are not a fan of this #cfg kani. It is true that it make the codebase a bit more complex but make the entire process of formal verification way easier. I lost a lot of time because in one test you have a precondition that you don't have in another and so on. **I really need** to put the #cfg kani to this place and a few csr registers in write CSR (like 3) such that I can merge everything and we remove this in the next commits.


CharlyCst:
Turns out we can just remove the `#[cfg]` now, so I simply did some cleanup :)

PR #394, opened 2025-01-28
Benchmark

francois141:


config/visionfive2-release-traps.toml
```diff
@@ -0,0 +1,25 @@
+# A simple configuration to run on StarFive VisionFive 2 platform in release mode
```
CharlyCst, line 1: We already have multiple VisionFive 2 configs, we need to clarify what this one does and why it is different from the others. It is not obvious from just reading the file for new people coming to the project.

src/benchmark/counter_per_code.rs
CharlyCst, line 1: What does "per code" mean? How are these counters different from the others? We need more context and comments to explain what this file is used for.

CharlyCst (changes requested):
Overall the PR looks good, but I am not convinced we need a new benchmark backend, I think the "counter" backend is enough, but we just need to make it per-core (as you did with the new file, so we could override the old one). The more we have the more maintenance is needed, so let's keep only what we need.

Is "per_code" a typo?

CharlyCst:
We also need better commit description: "This commit introduces a new benchmark used to debug and understand how the system behaves" does not explain how the new benchmark infrastructure is different and why we needed it. It only states what is obvious from the diff: we have more code now.
Overall commit descriptions need to explain why the project evolved the way it evolved.

francois141:
To clarify: it counts per mcause trap reason, not per core.

PR #398, opened 2025-01-30
Adjust sail preconditions and update model

francois141:
This commit adjust a few preconditions to run the formal verification, update the models and fixes at the same time a few bugs in the formal verification infrastructure.

Sry for the name of the branch; I wanted to assign you a little task, but then I realised it would make your life a bit easier if I give you the last two registers to check at the end of this pr wave :-)


CharlyCst:
@francois141 Just a small note:

I am not nitpicking too much, but ideally commit messages should convey the **why** and **how**, not the **what** which is already encoded in the diff.
For instance the original commit message was:

> Adjust sail preconditions and update model
> 
> This commit adjust a few preconditions to run the formal verification, update the models and fixes at the same time a few bugs in the formal verification infrastructure.

This doesn't convey much information, for instance:
- How was the model updated? What changed and why? The important information here is that we patched the upstream model with LCOFIE and we reflect the changes here.
- What is the "a few preconditions"? And "a few bugs"? There are changes regarding the general purpose registers, what was wrong? The key insight here is that in the general purpose register table we have an entry for x0 which should be hardwired at 0, hence your fixes. For a new developer on the project this is not obvious.
- Finally this commit bundles two things: LCOFIE in the model and a bugfix in the GP, they could be separate (even in the same PR, as long as the two commits are clean individually).

As an example I updated the commit with the relevant details.

PR #400, opened 2025-02-01
wip; preconditions

francois141:


francois141:
@CharlyCst Many preconditions fails, should we already merge it or fix the preconditions before and merge it later?

CharlyCst:
What kind of preconditions do you need? It seems there is a lot of them, I am wondering what the overlap is with the preconditions in the new symbolic context?

Do we have a clear idea of what should be fixed in the model checking, versus what needs to be encoded as a pre-condition in Miralis itself?

CharlyCst:
Closing for inactivity

PR #401, opened 2025-02-03
Explicit allow some illegal instructions

francois141:


CharlyCst (approved):
Thanks for the PR, and for the nice commit messages :)

PR #402, opened 2025-02-04
Introduce description for the Register enum

francois141:
In order to improve the code readability of the protect payload policy (and the rest of the codebase), this commits introduces docs comments in the Register enumeration, where each comment gives the abi name of the risc-v calling convention.


CharlyCst:
Thanks for the changes :)

PR #403, opened 2025-02-05
Clean code in policies

francois141:


CharlyCst (approved):
Thanks for the cleanup, looks good to me!

PR #406, opened 2025-02-10
Bugfix: compressed loads and stores have the wrong offset

francois141:
There is a bug in the decoding of compressed load and stores. The offset was always too big and this commit fixes to the correct offset. On top of that it fixes the wrong unit tests and add a todo! field in case we ever face stack-based compressed load and stores

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -11,6 +11,16 @@ const HFENCE_INSTR_GVMA_MASK: usize = 0b0110001 << 25;
 const RS1_RS1_INSTR_TYPE_MASK: usize = 0b1111111111000000001111111;
 const FUNC3_MASK: usize = 0b111000000000000;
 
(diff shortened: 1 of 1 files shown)


src/decoder.rs
```diff
@@ -1047,20 +1053,12 @@ mod tests {
...
-        // but we assert imm as 32 (8*4), as the eventual offset value would be 32.
-        // This helps to keep compressed and uncompressed instructions handlers more homogenous in other modules.
-
         assert_eq!(
             mctx.decode_store(0xffffe798),
             Instr::Store {
                 rs2: Register::X14,
                 rs1: Register::X15,
-                imm: 64,
+                imm: 8,
```
CharlyCst, line 1061: In short what was the issue @francois141 ? Why shouldn't we multiple the immediate to get the final offset?

PR #408, opened 2025-02-11
Introduce formal verification for compressed loads and stores

francois141:
This commit adds two new symbolic test for verifying the compressed loads and stores. The process found one bug which was fixed in a recent previous commit.


CharlyCst (commented):
Overall looks good to me, there is just one nit regarding the code organization: I am not a fan of creating two new crates when we already have one for the decoder. Could we create a single decoder crate instead, and have three modules there for the different sub-parts of the decoder? This also raise the question of how to name the "main" decoder, does it only decode CSRs? If yes is it the "CSR" or "privileged" decoder?

francois141:
Changes done

PR #410, opened 2025-02-13
Introduce offload policy and multiplex timer interrupts in Miralis

francois141:
This commit introduces a new policy. This policy is a bit different from the previous one in the sense that the main goal is the performance here. The previous measurements of Miralis showed that a huge amount of traps came from three sources. Reading the time register, setting the timers for the payload and emulating misaligned load and stores. This policy offloads the three points and handles the logic in Miralis.


src/device/clint.rs
```diff
@@ -1,18 +1,39 @@
...
 // ————————————————————————————— Virtual CLINT —————————————————————————————— //
 
+// We use this structure to avoid false sharing in the virtual clint
+// The typical size of a cache line is 64 bytes
+#[repr(C, align(64))]
+#[derive(Debug)]
+struct TimestampEntry {
+    counter: AtomicUsize,
+    _padding: [u8; 56],
+}
```
CharlyCst, line 26: If you are going for a struct with padding lets maybe add both timestamps (firmware and payload) in the same struct to limit bloating the code size. They are both read by the same core, so that will not add more cache contention while dividing the total timestamp size footprint by 2.

src/device/clint.rs
```diff
@@ -95,12 +168,39 @@ impl VirtClint {
...
+        self.write_clint(offset, w_width, value, ctx, false)
+    }
+
+    fn write_clint(
         &self,
         offset: usize,
         w_width: Width,
         value: usize,
         ctx: &mut VirtContext,
+        is_from_payload: bool,
```
CharlyCst, line 203: I would suggest creating a small enum, something like `ClintTimer::Payload` and `ClintTimer::Firmware`. It would be a bit more readable, but mostly I think we might eventually add a third one for Miralis itself (for instance if we want a policy that guarantees an enclave is going to run for X ms).

src/device/clint.rs
```diff
@@ -154,20 +254,43 @@ impl VirtClint {
...
+                        .counter
+                        .load(Ordering::SeqCst);
+                    let mtimecmp_payload = self.next_timestamp_payload[ctx.hart_id]
+                        .counter
+                        .load(Ordering::SeqCst);
+
+                    driver.write_mtimecmp(hart, min(mtimecmp_firmware, mtimecmp_payload))?;
+                }
+
+                self.propagate_payload_interupt_physically(ctx);
```
CharlyCst, line 289: We should do that only when updating the clint of behalf of the payload. The reason is that if we update the `mip.STIP` bit while the firmware is running then we will get and supervisor interrupt while running the firmware, which we definitely don't want :)

It might maybe not happen with the code right now, but let's be defensive and add an if there just to limit the risk of introducing bugs with future refactorings.

src/policy/offload.rs
```diff
@@ -0,0 +1,249 @@
...
+                            ctx.set(Register::try_from(rd).unwrap(), Arch::read_csr(Csr::Time));
+                            ctx.pc += 4;
+                            return PolicyHookResult::Overwrite;
+                        }
+                        0x1000 | 0x3000 | 0x5000 | 0x6000 | 0x7000 => {
+                            todo!("Handle the offload of other CSR instructions")
+                        }
+                        _ => {}
+                    }
+                }
```
CharlyCst, line 72: So we can't re-use part of the decoder here? The code seems very similar.

src/policy/offload.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    fn switch_from_firmware_to_payload(
+        &mut self,
+        _ctx: &mut VirtContext,
+        _mctx: &mut MiralisContext,
+    ) {
+    }
+
+    fn on_interrupt(&mut self, _ctx: &mut VirtContext, _mctx: &mut MiralisContext) {}
```
CharlyCst, line 102: Those could be default implementations, to remove the need for empty implementations in policy modules.

CharlyCst (commented):
Here are some feedback on the PR, thanks for the work on the offload :)

src/arch/metal.rs
```diff
@@ -282,7 +283,10 @@ impl Architecture for MetalArch {
             Csr::Mcycle => asm_read_csr!("mcycle"),
             Csr::Minstret => asm_read_csr!("minstret"),
             Csr::Cycle => todo!(),
-            Csr::Time => asm_read_csr!("time"),
+            Csr::Time => {
+                let clint = Plat::get_clint().lock();
+                value = clint.read_mtime();
+            }
```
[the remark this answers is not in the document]
francois141: Mtime is a memory mapped register at the location of the clint driver if I am not wrong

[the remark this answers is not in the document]
francois141: We should maybe kick out the lock in this scenario though

src/policy/offload.rs
francois141: Let's do it as a second step.

francois141:
Ready for second review

CharlyCst:
The PR needs quite a bit of cleanup, I'll handle it directly rather than doing a bunch of round-trip between the two of us.
You can focus on rebasing https://github.com/CharlyCst/miralis/pull/378 instead if you would like to see it merged before doing the benchmarks.

francois141:
What is the issue with this pr? I don't get it

francois141:
@CharlyCst When do you plan to be ready with this? Such that I know when I can start measuring ^^

CharlyCst:
Yup, it's over there: #416

I'm waiting for the test to pass, and I need to do a last check, but it's basically the same PR as this one with a bit of cleanup in the CLINT, and in the git history :)

CharlyCst:
Merged as part of #416 @francois141 :)

PR #411, opened 2025-02-13
Hide csr registers

francois141:


src/virt/mod.rs
```diff
@@ -233,6 +233,81 @@ pub struct VirtCsr {
 }
 
 impl VirtCsr {
+    pub fn new_empty() -> Self {
```
CharlyCst, line 236: You might already know it, but just in case here is a Rust tip:

When we need a default value for a struct (usually 0s or None for all fields) there is a trait [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) which we can derive, and then `VirtCsr::default()` will return exactly that.

In this particular case we can't add the `#[derive(Default)]` because the `Default` trait is implemented only for arrays up to 32 elements, but we have an array of 64 for the PMP addresses, so your implementation is the best we can do until Rust lift this limitation :)

CharlyCst (commented):
So far looks good to me :)

Did you have the chance to try it out on the board before merging?

src/virt/mod.rs
francois141: Thanks for letting me now. Indeed I was aware of that ^^

francois141:
It works on the board, merged as you approved it

PR #413, opened 2025-02-14
Average number of world switches and firmware exits by the number of …

francois141:
…cores

We are interested in the average number of world switches and firmware exits. Therefore we must average by the number of cores. On the visionfive 2 board, there is an edge case. One core doesn't boot linux and must therefore be excluded.

### DIFF
--- a/src/benchmark/counter.rs
+++ b/src/benchmark/counter.rs
@@ -5,6 +5,8 @@ use crate::benchmark::default::IntervalCounter;
 use crate::benchmark::Counter::{FirmwareExits, WorldSwitches};
 use crate::benchmark::{BenchmarkModule, Counter, Scope};
 use crate::config::PLATFORM_NB_HARTS;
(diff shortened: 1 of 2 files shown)


CharlyCst:
What if we run benchmarks on a single core then? We definitely want the exact number per core, not an average.

What we could do (if that helps), is to add a new parameter to choose which core to get the data from. Then from the kernel module we can simply iterate over the cores an collect all the results.

francois141:
There is already support for a single core and I don't want to iterate in the kernel module across all cores as this solution is easier to run with my current benchmark infrastructure as it display only a single log.

francois141:
We can get at any time the value per core, the goal of this pr is to get the average for the multicore benchmarks. You told me for the paper we want the average number of trap for multicore benchmarks

francois141:
After some careful consideration, I found a better solution. I will handle this in the benchmark infrastructure directly

PR #414, opened 2025-02-15
Deduplicate hardware extension availability in symbolic.rs

francois141:
We initialize twice the hardware extension in the formal verification, which could lead to inconsistencies. This commit introduces a quick refactoring to avoid consistency issues and subtle bugs.

### DIFF
--- a/model_checking/src/adapters.rs
+++ b/model_checking/src/adapters.rs
@@ -5,8 +5,9 @@
 //! implementation and Miralis we need to be able to compare their internal representation. Hence
 //! this module exposing functions to convert from one representation to the other.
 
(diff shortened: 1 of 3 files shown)


CharlyCst:
Indeed that sounds like a problem. The CI is failing with the changes, maybe it's because the values in `userspace` are different from the one we have in the model checking? If that's the cause, then copying the values from the model checking inside `userspace` should fix the Ci :)

francois141:
The pipelined failed because we had a third place where we were adding an extension capability. It wasn't related to the userspace. I fixed this as well.

francois141:
It was in sail_to_miralis

PR #415, opened 2025-02-17
Deduplicate hardware extension availability in symbolic.rs

francois141:
We initialize twice the hardware extension in the formal verification, which could lead to inconsistencies. This commit introduces a quick refactoring to avoid consistency issues and subtle bugs.


francois141:
Sry misclick for closing

CharlyCst:
Thanks for the changes :)

PR #417, opened 2025-02-18
Wip: splitting instr load and store

francois141:
@CharlyCst Would you mind ending this PR? I know some point required a bit of refactoring


CharlyCst:
@francois141 I merged the PR as #419.

I refactored a bit around the macros (we don't want to just use macros to duplicate code), and used the opportunity to clean-up some of our old code around load/store emulation using MPRV.

PR #418, opened 2025-02-18
Measure offload latency for timer

francois141:
The most two frequents operations that the payload requests from the firmware is to set the next supervisor timer interrupt and read the current time using the time register. This commit introduces a new benchmark to measure the latency of the operations in the offload policy.

@CharlyCst Feel free to review, I will create a second commit to insert in the ci / cd pipeline


CharlyCst (approved):

PR #421, opened 2025-02-18
Bugfix offload loads and stores

francois141:


CharlyCst:
For some reasons the test fails with a time out. I can reproduce the stall on my machine.

As we need to fix that for the measurements, we can start to merge the fix first. I fixed the root cause issue there: #422, can you check if it works for you? Rather than fixing that particular case in the load/store emulation I fixed the `read_raw_instr` function.

CharlyCst:
Closing the PR as the bugfix is merged in #422.

As we did not merge the test coverage due to timeout I opened an issue to track it #425.

PR #429, opened 2025-02-22
Boot benchmark

francois141:


CharlyCst (commented):
I cleaned-up a bit more regarding the removal of the old benchmark, as we had some more infrastructure around it.

I added back the OS firmware test, it doesn't depend on firmware and we definitely want to test it in the CI. 

Two other small notes:
- The code markers '`' are for inline code, whereas '```' are for multi-line code blocks. When you use code markers inline use a single back-tick, when you create a code block use three.
- Deprecated does not mean removed. Deprecating means that something will be removed in the future (and thus should not be used anymore for new code).

PR #432, opened 2025-02-23
Range over sfence vma

francois141:


CharlyCst:
I can reproduce the stall on my machine (which caused the CI to time-out) with:

```sh
just run linux config/test/qemu-virt-sstc-offload.toml
```

PR #433, opened 2025-02-25
Add building miralis for the `visionfive2` and `premierp550` boards i…

francois141:
…n the test pipeline

We won't detect all build errors in the `visionfive2` and `premierp550` in the platform at the moment. This commit adds them in the pipeline to be sure they are buildable after every commit.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,8 +1,10 @@
-default          := "default"
-config           := "config.toml"
-benchmark        := "ecall_benchmark"
-spike            := "./config/spike.toml"
-qemu_virt        := "./config/test/qemu-virt.toml"
+default      := "default"
+config       := "config.toml"
+benchmark    := "ecall_benchmark"
+spike        := "./config/spike.toml"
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -32,6 +34,9 @@ test:
 
 	# Test firmware build
 	just build-firmware default {{qemu_virt}}
+	just build-firmware default {{visionfive_2_build}}
+	just build-firmware default {{premier_p550_build}}
```
CharlyCst, line 38: So this tests building the `default` firmware for the given boards, not Miralis itself.

To test the build of Miralis we need to run `just build {{premier_p550_build}}` instead.

As a note, this will add ~11s per board (on my machine). It is quite a lot, but I agree we should test it still.
The long build time is due to the re-compilation of the crates (`tiny-keccak` is the longest to build for me), maybe we can find a way to prevent re-compilation of dependencies at least.

francois141: Oops, this is the problem when you work too late ^^

Thanks for spotting it!

PR #434, opened 2025-02-25
Refactor clint

francois141:


src/device/clint.rs
```diff
@@ -54,8 +51,6 @@ impl TimestampEntry {
 /// Represents a virtual CLINT (Core Local Interruptor) device
 #[derive(Debug)]
 pub struct VirtClint {
-    /// A driver for the physical CLINT
-    driver: &'static Mutex<ClintDriver>,
```
CharlyCst, line 58: We do want to keep a driver in the virt clint. There might be multiple CLINT (e.g. on NUMA systems).
What we want to do is remove the lock.

src/device/clint.rs
```diff
@@ -89,9 +84,8 @@ impl DeviceAccess for VirtClint {
 
 impl VirtClint {
     /// Creates a new virtual CLINT device backed by a physical CLINT.
-    pub const fn new(driver: &'static Mutex<ClintDriver>) -> Self {
```
CharlyCst, line 92: Keep the driver, drop the lock.

src/device/clint.rs
```diff
@@ -115,10 +109,7 @@ impl VirtClint {
         let timestamps = &self.next_timestamps[mctx.hw.hart];
 
         // Read current timestamp and drop the lock right away
-        let current_timestamp: usize = Plat::get_clint()
```
CharlyCst, line 118: should be `self.driver`, and no lock afterward.

src/device/clint.rs
```diff
@@ -159,30 +150,26 @@ impl VirtClint {
         let next_deadline = min(firmware_deadline, payload_deadline);
 
         // Write the next deadline back
-        Plat::get_clint()
```
CharlyCst, line 162: Same, should be `self.driver`.

src/device/clint.rs
```diff
@@ -159,30 +150,26 @@ impl VirtClint {
...
-                driver.read_mtimecmp(hart)
+                clint_driver::read_mtimecmp(hart)
             }
-            (o, Width::Byte8) if o == MTIME_OFFSET => Ok(driver.read_mtime()),
+            (o, Width::Byte8) if o == MTIME_OFFSET => Ok(clint_driver::read_mtime()),
             // We also handle the case of 4 bytes reads to mtime
-            (o, Width::Byte4) if o == MTIME_OFFSET => Ok(driver.read_mtime() & 0xffffffff),
-            (o, Width::Byte4) if o == MTIME_OFFSET + 4 => Ok(driver.read_mtime() >> 32),
+            (o, Width::Byte4) if o == MTIME_OFFSET => Ok(clint_driver::read_mtime() & 0xffffffff),
+            (o, Width::Byte4) if o == MTIME_OFFSET + 4 => Ok(clint_driver::read_mtime() >> 32),
```
CharlyCst, line 172: We should replace the `clint_driver` by `self.driver`, same in the following.

src/driver/clint.rs
```diff
@@ -18,32 +14,32 @@ pub const MSIP_WIDTH: Width = Width::Byte4;
 pub const MTIMECMP_WIDTH: Width = Width::Byte8;
 pub const _MTIME_WIDTH: Width = Width::Byte8;
 
-#[derive(Clone, Debug)]
-pub struct ClintDriver {
-    /// The base address of the physical CLINT.
-    base: usize,
-}
+pub mod clint_driver {
```
CharlyCst, line 17: We do not want to have a single global driver. In general, any kind of "ambient authority" is often bad design and ends up biting back in the long term.
Here a typical issue would be supporting NUMA nodes with two CLINTs.
Another issue would be determining dynamically the driver we should use (e.g. if we want to support ACLINT features, as we have one the P550).

So in short, the previous architecture is already functional, the only modifications needed is to remove the `Mutex`and pass `&self` instead of `&mut self` as arguments.

src/driver/clint.rs
```diff
@@ -18,32 +14,32 @@ pub const MSIP_WIDTH: Width = Width::Byte4;
 pub const MTIMECMP_WIDTH: Width = Width::Byte8;
 pub const _MTIME_WIDTH: Width = Width::Byte8;
 
-#[derive(Clone, Debug)]
-pub struct ClintDriver {
-    /// The base address of the physical CLINT.
-    base: usize,
-}
```
CharlyCst, line 25: We want to keep this structure.

Also a note on performance, it might seem faster to not have a `base` field se we can use a constant. Turns out it is not necessarily the case. On RISC architecture loading a constant is expensive in terms of instructions, because instructions are 4 bytes and can only load a few bits at a time, so loading a constant ends-up taking ~4 instructions, whereas loading the value from a struct takes 1. Of course, there are caches, so the only way to know which is faster is to benchmark. I am just raising the point that's is not necessarily a bad thing to keep the base in the struct performance-wise, and it gives us more flexibility.

src/platform/miralis.rs
```diff
@@ -4,46 +4,25 @@ use core::fmt;
...
 use crate::device::VirtDevice;
-use crate::driver::clint::ClintDriver;
+use crate::platform::{Plat, VIRT_CLINT, VIRT_TEST_DEVICE};
 use crate::Platform;
 
-// —————————————————————————— Platform Parameters ——————————————————————————— //
-
-const MIRALIS_START_ADDR: usize = TARGET_FIRMWARE_ADDRESS;
-const FIRMWARE_START_ADDR: usize = TARGET_PAYLOAD_ADDRESS;
-const CLINT_BASE: usize = 0x2000000;
```
CharlyCst, line 20: We can factor-our the MIRALIS_START and FIRMWARE_START, but let's keep the clint address in the platform folder. It is clearer to keep all platform variables in their own file, even though there are some standard addressees for most platforms.

CharlyCst (commented):
So two things here: there is a refactor of the clint driver and one for the device platform. We should deal with those two separately in two commits.

- Regarding the clint, we want to keep the old way (i.e. using a clint driver), but remove the lock. The old architecture was design for extensibility, we should not roll back that part (see comments).
- Regarding the platform refactor, I agree we should do some clearnup. Let's just keep the CLINT and UART addresses within the platform file.

CharlyCst:
@francois141 I'll handle the PR, keeping your changes on the platform but re-factoring the CLINT using the existing code base.

PR #440, opened 2025-03-18
Forward to smode directly

francois141:


src/policy/offload.rs
```diff
@@ -212,4 +221,142 @@ impl OffloadPolicy {
...
+        assert!(
+            !prev_virt,
+            "Currently, we never tested this code when virtualisation is active, the feature might be unstable"
+        );
+
+        // By default, we redirect to HS-mode
+        let mut next_virt: bool = false;
+
+        // Sanity check on previous mode
+        let prev_mode = (mstatus & MPP_FILTER) >> MPP_OFFSET;
```
CharlyCst, line 244: We have a function to get the mode, its `parse_mpp_return_mode`. That will also remove the need for the "as usize" later on.

src/policy/offload.rs
```diff
@@ -212,4 +221,142 @@ impl OffloadPolicy {
...
+        if prev_mode != Mode::S as usize && prev_mode != Mode::U as usize {
+            unreachable!()
+        }
```
CharlyCst, line 247: This should be an assert

src/policy/offload.rs
```diff
@@ -212,4 +221,142 @@ impl OffloadPolicy {
...
+
+    /// This function is a rust implementation of the function "sbi_trap_redirect" in the sbi_trap.c from the OpenSBI codebase
+    /// This function corresponds to a payload to payload transition. Therefore, we don't modify the virtual context but the physical registers
+    /// The only exception is ctx.pc because in the trap handler we write mepc with ctx.pc value.
+    #[allow(unused)]
+    fn redirect_trap_to_s_mode(ctx: &mut VirtContext) -> PolicyHookResult {
+        let mut mstatus = ctx.trap_info.mstatus;
+
+        // The previous virtualisation mode
+        let prev_virt: bool = mstatus & MPV_FILTER != 0;
```
CharlyCst, line 233: Small nit:

For booleans a good naming convention is to have a `is` somewhere, such as `is_prev_virt` or `prev_is_virt`. That makes it easier to identify boolean variables, and also makes conditionals more readable, like `is prev_is_virt {}`.

CharlyCst (changes requested):

CharlyCst (approved):

PR #462, opened 2025-07-14
Add 'path' config to QEMU binary

danielhb:
Adding an optional 'path' option to QEMU allows to run different qemu-system-riscv64 flavors/builds
by just changing config.toml, instead of having to tinker with the PATH env var.

This change was done to help debug and solve QEMU bug https://gitlab.com/qemu-project/qemu/-/issues/3020. I created a config.toml file like this:

```
$ cat config.toml 
[log]
level = "trace"

[qemu]
path = "/home/danielhb/work/qemu/build"
```

And I was able to run the qemu-system-riscv64 binary from specified path.


CharlyCst (approved):

danielhb:
Sorry for not paying attention to the CI results. It should be ok now

